PR #3, opened 2016-06-24
Kill java

nacho:
Maintaining binding on the same module is a pain, at the end
you always end up fighting with distcheck, if somebody want
to put this into its own cyrus-sasl-java module I'd be very
glad but I really prefer to take it out of the way from the main
module.


quanah:
@ksmurchison What are your thoughts on this?  The java module seems completely unmaintained, I question if it would even work with modern OpenJDK.  Additionally Java has its own SASL stack now, not sure why this code is necessary at this point?

Thanks!

ksmurchison:
Seems fine for 2.2

quanah:
Closing in favor of up to date work done in #597

PR #5, opened 2016-08-25
clean up ndbm backend for sasldb

obache:
Just change to same as other backends.


nacho:
Patches look good to me

PR #10, opened 2016-12-01
Missing definition of OPENSSL_zalloc for old OpenSSL versions

Jakuje:


saslauthd/lak.c
```diff
@@ -1716,6 +1716,15 @@ static int lak_check_password(
 #define EVP_MD_CTX_new()      EVP_MD_CTX_create()
 #define EVP_MD_CTX_free(ctx)  EVP_MD_CTX_destroy((ctx))
 
+static void *OPENSSL_zalloc(size_t num)
```
nacho: do we need to ifdef this? I guess this would be redefined with new versions of openssl?

Jakuje: There is `ifdef` already above on the line 1715: ` #if OPENSSL_VERSION_NUMBER < 0x10100000L`

nacho:
OK, I couldn't see it on the diff. For me it is fine to get this in.

PR #420, opened 2017-02-15
Fix authdaemond authentication on OpenBSD

bigio:
select(2) returns the total number of ready descriptors in all the sets which can be more than 1, this fixes courier-authdaemond sasl authentication on, at least, OpenBSD.


Jakuje:
The SPNEGO changes should probably not go in. Try to rebase (instead of merging) your changes against the upstream to avoid confusion.

bigio:
Hope I do it right, otherwise I will start from scratch.

PR #423, opened 2017-02-22
document error in install.html

yu2003w:
After check code out from git, need to run 'sh ./autogen.sh' to generate necessary configurations.


nacho:
Looks good.

yu2003w:
@nacho Thanks for review.

Issue #427, opened 2017-03-28
wrong ldap binding in group check


slesru:
Hi!

From here:

https://bugs.launchpad.net/ubuntu/+source/cyrus-sasl2/+bug/1188475

Problem is that after user is authentificated with ldap bind , ldap
connection for checking user in group ( lak_group_member function )
is made with this user's bind, not bind parameters from config file.
User can not ( and have not in our case- I don't know why , but this is
not real problem ) have access to ldap groups.
And so, authentication is always fail.

Patch:

 diff -ur lak.c.orig lak.c
--- lak.c.orig 2013-06-07 09:15:20.098788278 +0400
+++ lak.c 2013-06-08 10:17:07.548233104 +0400
@@ -1448,8 +1448,25 @@
     if ( rc == LAK_OK &&
         (ISSET(lak->conf->group_dn) ||
          ISSET(lak->conf->group_filter)) )
- rc = lak_group_member(lak, user, service, realm, dn->value);
+ {
+ /* restore config bind */
+ lak_unbind(lak);
+ rc = lak_user(
+ lak->conf->bind_dn,
+ lak->conf->id,
+ lak->conf->authz_id,
+ lak->conf->mech,
+ lak->conf->realm,
+ lak->conf->password,
+ &lu);
+ if (rc != LAK_OK)
+ goto done;
+ rc = lak_bind(lak, lu);
+ if (rc != LAK_OK)
+ goto done;

+ rc = lak_group_member(lak, user, service, realm, dn->value);
+ }
 done:;
     if (lu)
         lak_user_free(lu);

Still not fixed...

ksmurchison:
Fixed with https://github.com/cyrusimap/cyrus-sasl/commit/f4d4f5121a84829353697556f5d83467e9cf16f5

PR #428, opened 2017-03-28
Update lak.c

slesru:
fixes old bug
https://bugs.launchpad.net/ubuntu/+source/cyrus-sasl2/+bug/1188475

https://github.com/cyrusimap/cyrus-sasl/issues/427


Jakuje:
What is the point of the `<>` at the end of the line with `lak_user()`? Is it typo? This change breaks my build.

slesru:
Yes, this was a typo.
Thank you!

Issue #372, opened 2012-08-17
A bug in sasl_gss_encode can cause SEGV crash


brong:
From: [Geoff Adams](gadams+cyrus-bugzilla@avernus.com)
Bugzilla-Id: 3732
Version: 2.1.25
Owner: [Alexey Melnikov](Alexey.Melnikov@isode.com)

quanah:
This bug needs to be reopened, and caused continued SEGVs

quanah:
Above commit from debian patchset:

```
 cat 0033_fix_segfault_in_GSSAPI.patch
--- cyrus-sasl2.orig/plugins/gssapi.c
+++ cyrus-sasl2/plugins/gssapi.c
@@ -376,6 +376,7 @@ sasl_gss_encode(void *context, const str
     }

     if (output_token->value && output) {
+       int len;
        unsigned char * p;

        ret = _plug_buf_alloc(text->utils,
@@ -392,11 +393,8 @@ sasl_gss_encode(void *context, const str

        p = (unsigned char *) text->encode_buf;

-       p[0] = (output_token->length>>24) & 0xFF;
-       p[1] = (output_token->length>>16) & 0xFF;
-       p[2] = (output_token->length>>8) & 0xFF;
-       p[3] = output_token->length & 0xFF;
-
+       len = htonl(output_token->length);
+       memcpy(text->encode_buf, &len, 4);
        memcpy(text->encode_buf + 4, output_token->value, output_token->length);
     }

```

Issue #340, opened 2011-04-18
plugins/gssapi.c segfaults in sasl_gss_encode (CVS only)


brong:
From: [Dan White](dwhite@olp.net)
Bugzilla-Id: 3445
Version: 2.1.x
Owner: [Alexey Melnikov](Alexey.Melnikov@isode.com)

brong:
Closing: was RESOLVED in import

PR #442, opened 2017-07-14
For https://github.com/cyrusimap/cyrus-sasl/issues/372 Fix segfault w…

quanah:
…ith GSSAPI mech

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -399,6 +399,7 @@ sasl_gss_encode(void *context, const struct iovec *invec, unsigned numiov,
     }
     
     if (output_token->value && output) {
+	int len;
 	unsigned char * p;
 	
 	ret = _plug_buf_alloc(text->utils,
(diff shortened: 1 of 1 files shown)


ksmurchison:
Is the original code assuming a particular endianess which fails on some machines?

quanah:
This comes from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=629589

quanah:
Which in github, refers back to #372 and #340

plugins/gssapi.c
```diff
@@ -399,6 +399,7 @@ sasl_gss_encode(void *context, const struct iovec *invec, unsigned numiov,
     }
     
     if (output_token->value && output) {
+	int len;
```
aamelnikov: I think this be uint32_t, int is really icky here. I think int can still be 2 bytes on some platforms.

Issue #318, opened 2009-07-23
SASL man page should reference web documentation


brong:
From: [David Gatwood](dgatwood@mac.com)
Bugzilla-Id: 3168
Version: 2.1.x
Owner: [Ken Murchison](murch@andrew.cmu.edu)

PR #450, opened 2017-07-18
[Docs] Converting man pages to rst.

nicolan:
conf.py automatically produces the list of man pages.

[Docs] Link to online sysadmin guide from sasl.3

GitHub issue #318

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -43,7 +43,7 @@ extensions = [
 extensions.append('sphinxlocal.builders.manpage')
 extensions.append('sphinxlocal.roles.saslman')
 
-intersphinx_mapping = {'cyrusimap': ('http://www.cyrusimap.org/dev', None)}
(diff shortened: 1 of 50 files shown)


docsrc/conf.py
```diff
@@ -268,9 +269,50 @@
...
 
+# For each man page section/path,
+# look for all the rst files.
+# With each file, check if there's an '.. author: ' attribution (strict on
+# spacing and case sensitivity) and add to the author info.
+# Then add the file with all its details into the man_page array.
+# If the file is an :orphan:, then don't include it.
+current = os.path.abspath(os.getcwd())
+for tuple in pathset:
+    os.chdir(tuple[0])
```
elliefm: This line errors out for me because `sasl/reference/manpages/systemcommands/` doesn't exist.

I think it would be good to make it just skip that tuple if the directory doesn't exist?  That way if we later add those types of manpages, it'll already know how to find them.

I assume but haven't verified that if I remove systemcommands from pathset, then it will choke on usercommands or configs (which also don't currently exist)...

PR #462, opened 2017-08-01
initialise pluglist and plug_entry with NULL

ayoubserti:
Initializing pluglist and plug_entry with NULL pointer To avoid unpredictable errors.

Also:
When libsasl is compiled as static library using VS2013+ it cause a link error ( C4703 ) .

This pull request comes to solve thoses problems.


midrissi:
bitmoji

PR #468, opened 2017-09-02
add a ldapsimple pwcheck plugin

comicfans:
this mainly intend to be used under windows without permission to config
ldap server . for example setting up a svnserve on windows, but don't have
 admin account for domain controller. no config needs to be done on ldap server side,
but have following limitations:

all user must under same DN
user DN must be written in config
not support admin bind-search then bind process

sample config:
pwcheck_method: ldapsimple
mech_list: PLAIN
ldapsimple_servers:  ldap.forumsys.com
ldapsimple_dn: cn=%u,dc=example,dc=com
ldapsimple_port:389

currently this prototype only written and tested on windows, on other system you can use saslauthd to support full ldap authorization.

### DIFF
--- a/lib/checkpw.c
+++ b/lib/checkpw.c
@@ -95,6 +95,7 @@
 # endif
 #endif
 
+# include <winldap.h>
 
 /* we store the following secret to check plaintext passwords:
  *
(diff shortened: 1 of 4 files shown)


lib/checkpw.c
```diff
@@ -95,6 +95,7 @@
 # endif
 #endif
 
+# include <winldap.h>
```
aamelnikov: This include clearly breaks Unix platforms. So you need to wrap it in some form of #ifdef

comicfans: should  _WIN32 macro test enough ?  if so I'll update this

quanah:
@comicfans Unfortunately someone pushed master into your merge request, making it a complete mess.  Can you either update this MR so it's correct, or create a new one with just your changes?  Thanks!

It will need to be signed off on to comply with the DCO check as well.

Neustradamus:
@comicfans: Have you seen the @quanah comment?

comicfans:
> @comicfans: Have you seen the @quanah comment?
Hi @quanah @Neustradamus sorry for the late reply, this MR opened at 2017, because my working environment needs exactly this function, and I tested/used this function for a while. but currently I've changed my job (at 2018) so I don't have such testing environment, also lose the usage intention. so if possible, please close this. unfortunately, I don't have too much sparse time for this.

williamdes:
Can this PR be rebased so some testing can be done ?
I am interested in this plugin

hyc:
You can examine the code without all the extraneous commits on the original branch https://github.com/comicfans/cyrus-sasl/tree/fix-msc-1900

I note that the ldap_init / ldap_connect APIs are deprecated, it should be using ldap_initialize() instead.

Pfiver:
I'd love this kind of plugin available.

But I see a potential security issue in this implementation. (See https://github.com/comicfans/cyrus-sasl/commit/639f9832ff4b3424b29785b97565343c95bd44e9#r115512817)

I also found a similar Idea (patch) from a mailing list  contribution in 2001 (!) and am really wondering why that never made it to the trunk. https://www.openldap.org/lists/openldap-software/200108/msg00075.html

Use case: I also have an openLDAP instance where I store hashed passwords - thus I can not use the "auxprop" plugins. Or am I missing some other obvious solution?

quanah:
Needs DCO and merge conflict updating. May be easier to close this PR and create a new one.

Neustradamus:
@comicfans: Any progress on this PR?

comicfans:
> @comicfans: Any progress on this PR?

Hi @Neustradamus please feel free to make any changes to fix the issue, I'm not working on this topic anymore, sorry for that

Neustradamus:
@comicfans: 8 years after the ticket creation, I understand for this closing but badly it has not been merged :/
Important: the team has requested you to add the DCO:
- https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin
- https://www.google.com/search?q=dco+code+sign

PR #482, opened 2017-10-10
Prevent error messages being overwritten

four2five:
Update the SETERROR macro to test whether an
error message has already been set for a
connection prior to calling seterror.


four2five:
This patch is intended to further the ideal stated in this line from the docs

> In order to ensure consistency of this information, it is the responsibility of the deepest function with access to the sasl_conn_t make the call to set the errdetail string

ksmurchison:
This looks sane to me.  @aamelnikov does this look OK to you?

common/plugin_common.h
```diff
@@ -109,9 +109,12 @@ PLUG_API int sasl_canonuser_init(const sasl_utils_t *utils, \
...
 }
 
-/* note: msg cannot include additional variables, so if you want to
+/* Only set the error message if one is not already set. This prevents overwriting
+ * plugin-supplied error messages with library error messages.
+ * note: msg cannot include additional variables, so if you want to
  * do a printf-format string, then you need to call seterror yourself */
-#define SETERROR( utils, msg ) (utils)->seterror( (utils)->conn, 0, (msg) )
+#define SETERROR( utils, msg ) \
+    if (!sasl_errdetail((utils)->conn)) (utils)->seterror( (utils)->conn, 0, (msg) )
```
aamelnikov: I am worried that sasl_errdetail() returns non NULL value whether or not the error was set on the connection. I need to test whether sasl_errdetail() is the right thing here, it might not be.

four2five: I'm very new to SASL, so apologies if this is obvious question, but a freshly initiliazed server context returning a non-NULL value via sasl_errdetail() would not be inline with the documented return values: https://github.com/cyrusimap/cyrus-sasl/blob/master/docsrc/sasl/reference/manpages/library/sasl_errdetail.rst#return-value

That doesn't mean that it isn't happening, but it does raise a separate concern.

I'll work on adding some test cases to this patch and in doing so, I will address the concern that you've raised.

four2five:
I have no intention of working on this patch going forward (this was a tangential concern of mine two jobs ago). Inclined to close it unless someone else wants to pick it up.

quanah:
Please leave it open, thanks.

PR #497, opened 2017-12-26
Fallback gracefully for missing GSS_C_SEC_CONTEXT_SASL_SSF support in Heimdal

whitehse:
This addresses an issue introduced by commit ff9f9caeb6db6d7513128fff9321f9bd445f58b7, which breaks Heimdal backed GSSAPI authentication.

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -702,6 +702,9 @@ static int gssapi_get_ssf(context_t *text, sasl_ssf_t *mech_ssf)
     case GSS_S_UNAVAILABLE:
         /* Not supported by the library, fallback to default */
         goto fallback;
+    case GSS_S_FAILURE:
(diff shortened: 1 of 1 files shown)


simo5:
I am not sure this commit is ok.
A failure can happen in a library supporting ssf, and the only way to report it is via GSS_S_FAILURE. This patch treats it as a non-error.
If Heimdal is reporting  failure when a function is not unavailable I think Heimdal should be fixed.

However at the very least the minro should be checked to see if it is an error to skip or not. Heimdal returns a minor of 0 on failure, if a minor != 0 is found an error should probably be reported instead.

@whitehse can you please add checking for non 0 minor ?

whitehse:
I confirmed minor == 0 in this condition. I'll work up a patch.

PR #503, opened 2018-02-01
fix snprintf macro

mehdadi:


mehdadi:
compile using Visual studio 2017 does not need this macro. 
this patch will fix the conflict between current macro and the one defined in stdio.h in sdk 10.0.16...

nacho:
Patch looks good to me.

PR #510, opened 2018-04-13
Correct of signedness for proper url encoding variables with idn chars.

kwiatek6324:


ksmurchison:
'in' accesses elements of 'string' which is 'char *', so I don't see any mismatch here.

kwiatek6324:
Please try use with testsaslauthd password with any utf-8 international char (we have problem with polish letters "ąłóźżćó"). In this case in original code there are errors with url encoded parameters to web server. After provided patch everything is ok.

PR #512, opened 2018-05-11
configure.ac: fix condition for suppliment snprintf implementation

obache:
$sasl_cv_snprintf means requremnt of suppliment snprintf
implementation, not existence of system snprintf implementation,


ksmurchison:
snprintf() failed to compile on my machine.  Reverted this for 2.1.27

obache:
>           snprintf() failed to compile on my machine.  Reverted this for 2.1.27

It is curious.
If "compile snprintf()" means your machine is missing snprinf()  and need to build  bundled snprintf.c,
then "snprintf() failed to compile" means it is failed to resolve missing snprintf().

so after reverted,
your machine is missing snprintf(), but not compile snprintf.o, so it means no snprintf() implementation.

PR #533, opened 2018-09-20
Add MIT Kerberos as GSSAPI provider

marcodiiga:
Add support for MIT Kerberos 64 bit GSSAPI provider


plugins/NTMakefile
```diff
@@ -122,6 +122,9 @@ OPENSSL_FLAGS=
 !IF "$(GSSAPI)" == "CyberSafe"
 GSS_FLAGS= /I $(GSSAPI_INCLUDE) /D "HAVE_GSS_C_NT_HOSTBASED_SERVICE" /D "HAVE_GSS_C_NT_USER_NAME"
 GSS_LIBS=/libpath:$(GSSAPI_LIBPATH) gssapi32.lib
+!ELSE IF "$(GSSAPI)" == "MITKerberos"
+GSS_FLAGS= /I $(GSSAPI_INCLUDE) /D "HAVE_GSS_C_NT_HOSTBASED_SERVICE" /D "HAVE_GSS_C_NT_USER_NAME"
+GSS_LIBS=/libpath:$(GSSAPI_LIBPATH) gssapi64.lib
```
nacho: I think we need another variable where we can specify whether we want gssapi64 or gssapi32

marcodiiga: Done, added a compile option for that

nacho:
@ksmurchison this looks good to me, can we merge it before the release?

ksmurchison:
Yes, but I am having trouble building SASL right now, either because I don't have the proper tooling to build the docs, or the doc build relies on legacy stuff.  I;m trying to resolve this.

PR #535, opened 2018-09-24
Convert obsolete HTML references to rst form.

de0u:
Files: sysadmin.rst, components.rst, quickstart.rst


de0u:
The web documentation for saslauthd has some broken links.  If this isn't the right way to get them fixed, please let me know what would be a better approach.  Thanks!

Novaliss:
Good example link. 😉❤️

PR #536, opened 2018-10-05
dlopen: on osx the plugins are built as .so

nacho:
I have no idea why at some point it was changed to .plugin
but autotools defaults to .so and we are actually getting
.so files built.


nacho:
@ksmurchison we should definitely get this in before the release or the next rc

PR #537, opened 2018-10-09
Coverity again

Jakuje:
The coverity scan reported few more issues and the attached patch set is attempting to satisfy them.

One looks like obvious copy&paste error, then there is unhandled `va_end()` in case of error in two places and the other is attempt `strncpy` handling in the code. Unfortunately, the `strlcpy()` is complied only into the saslauthd so it can not be used outside of that.


Jakuje:
I went through the whole list and fixed few more low-severity issues reported by coverity. Additionally, there is bunch of the incompatible function types reported by compiler, that I was not sure how to handle:
```
saslpasswd.c:256:23: warning: cast between incompatible function types from ‘int (*)(void *, const char *, const char *, const char **, unsigned int *)’ to ‘int (*)(void)’ [-Wcast-function-type]
     { SASL_CB_GETOPT, (sasl_callback_ft)&good_getopt, NULL },
                       ^
```

lib/auxprop.c
```diff
@@ -390,6 +390,10 @@ void prop_clear(struct propctx *ctx, int requests)
     /* We're going to need a new proppool once we reset things */
     new_pool = alloc_proppool(ctx->mem_base->size +
 			      (ctx->used_values+1) * sizeof(struct propval));
+    if (new_pool == NULL) {
+	_sasl_log(NULL, SASL_LOG_ERR, "failed to allocate memory\n");
```
nacho: fix indent here

lib/checkpw.c
```diff
@@ -786,7 +790,8 @@ static int saslauthd_verify_password(sasl_conn_t *conn,
 
     memset((char *)&srvaddr, 0, sizeof(srvaddr));
     srvaddr.sun_family = AF_UNIX;
-    strncpy(srvaddr.sun_path, pwpath, sizeof(srvaddr.sun_path));
+    strncpy(srvaddr.sun_path, pwpath, sizeof(srvaddr.sun_path) - 1);
+    srvaddr.sun_path[strlen(pwpath)] = '\0';
```
nacho: shouldn't this be strlen - 1?

lib/dlopen.c
```diff
@@ -270,6 +270,8 @@ static int _parse_la(const char *prefix, const char *in, char *out)
     if (strcmp(in + (length - strlen(LA_SUFFIX)), LA_SUFFIX)) {
 	if(!strcmp(in + (length - strlen(SO_SUFFIX)),SO_SUFFIX)) {
 	    /* check for a .la file */
+            if (strlen(prefix) + strlen(in) + strlen(LA_SUFFIX) + 1 >= MAX_LINE)
```
nacho: wrong indent?

plugins/scram.c
```diff
@@ -1538,6 +1537,10 @@ static int scram_server_mech_step(void *conn_context,
 	return SASL_BADPROT;
     }
 
+    scram_sasl_mech = (EVP_MD_size(text->md) == 32)
+        ? "SCRAM-SHA-256"
```
nacho: I think ? should be on the previous line

utils/dbconverter-2.c
```diff
@@ -334,7 +334,10 @@ int listusers(const char *path, listcb_t *cb)
     }
 
     result = cursor->c_close(cursor);
-    if (result!=0) result = SASL_FAIL;
+    if (result!=0) {
```
nacho: take the opportunity and fix the style, add space before and after !=

plugins/scram.c
Jakuje: OK, I will revert to the original look. It is probably matter of personal preference.

lib/auxprop.c
Jakuje: Fixed to match surroundings

Jakuje:
@nacho Thank you for the review. I addressed your comments and pushed a new version.

simo5:
@nacho is this good to be merged yet, or does it need more review ?

PR #540, opened 2018-10-14
Add missing string.h include for strcmp().

hannob:
The file
pwcheck/pwcheck_getspnam.c
contains a call to strcmp(), but does not include string.h.

While this will sometimes work due to indirect includes of string.h or lenient compilers, it's not correct and you cannot rely on it. Therefore it's better to explicitly include headers for the functions used.

(I found this while going through the patches in the Gentoo package of cyrus-sasl, and this seemed obvious and simple enough to push upstream.)

### DIFF
--- a/pwcheck/pwcheck_getspnam.c
+++ b/pwcheck/pwcheck_getspnam.c
@@ -42,6 +42,7 @@
  */
 
 #include <shadow.h>
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


nacho:
Sounds good to me, @ksmurchison simple enough to get in.

okias:
can be merged? :+1:

PR #542, opened 2018-10-29
Emit debug logging only in case of errors

simo5:
This kind f debug statement is barely useful when there are errors.
Emit it only on error, not on every normal iteration.
This cuts down on useless logging during normal operations while
still giving some quick hint in case of trouble.


simo5:
@Jakuje please take a look

Jakuje:
Logging the action before something is done is helpful for debugging in case of the application crashes or goes into unexpected branches after that. I hope that is no longer common here. Indeed I agree that for other use cases we should not log it, especially if there is no easy way to suppress them.

Would it make sense to add some error information (code or whatever is in the gssapi used on failure) to the log message once we are in modifying the logs?

simo5:
It seems actual gss error messages are already logged via appropriate loggers, that is why I just moved the debug message and haven't added them.

plugins/gssapi.c
```diff
@@ -1488,8 +1485,10 @@ gssapi_server_mech_step(void *conn_context,
 	}
 
 	oparams->doneflag = 1;
+    } else {
+        params->utils->log(params->utils->conn, SASL_LOG_DEBUG,
+		           "GSSAPI server step failed: %d\n", text->state);
     }
-
```
nacho: keep the empty line before the return

simo5: will do

quanah:
@simo5 Needs re-push to address DCO sign off
Needs empty line issue raised by @nacho addressed, thanks!

simo5:
done

hyc (approved):

PR #554, opened 2019-02-02
db_gdbm: fix gdbm_errno overlay from gdbm_close

TheJJ:
`gdbm_close` also sets gdbm_errno since version 1.17.
This leads to a problem in `libsasl` as the `gdbm_close` incovation overlays
the `gdbm_errno` value which is then later used for the error handling.

Downstream bugs:
* Archlinux: https://bugs.archlinux.org/task/59873

### DIFF
--- a/sasldb/db_gdbm.c
+++ b/sasldb/db_gdbm.c
@@ -107,9 +107,11 @@ int _sasldb_getdata(const sasl_utils_t *utils,
   gkey.dptr = key;
   gkey.dsize = key_len;
   gvalue = gdbm_fetch(db, gkey);
+  int fetch_errno = gdbm_errno;
+
   gdbm_close(db);
   if (! gvalue.dptr) {
-      if (gdbm_errno == GDBM_ITEM_NOT_FOUND) {
+      if (fetch_errno == GDBM_ITEM_NOT_FOUND) {
           utils->seterror(conn, SASL_NOLOG,
 			  "user: %s@%s property: %s not found in %s",
 			  authid, realm, propName, path);


TheJJ:
Would you please merge this fix?

hhoffstaette:
This bug prevents users from moving to a berkeleydb-free system, which is highly desirable due to bdb's licensing and QA status. The patch works fine, *please* merge it.

quanah:
I would note that current cyrus-sasl supports LMDB as an alternative to BerkeleyDB.  So it's not clear to me how this prevents one from going completely BerkeleyDB free?

hhoffstaette:
> I would note that current cyrus-sasl supports LMDB as an alternative to BerkeleyDB.

Thanks, that's good to know - my package on Gentoo doesn't offer lmdb as option, so it didn't occur to me to try (probably just an oversight, easily corrected). I confirmed that it works when running
configure manually, so that's a good option since I already use lmdb in postfix as well.

hhoffstaette:
> I would note that current cyrus-sasl supports LMDB as an alternative to BerkeleyDB.

Have you actually tested it? It seems to suffer from the same problem as gdbm without the patch:
failing auth despite cleanly recreated sasldb2 with valid credentials.
Back to gdbm..

quanah:
@hhoffstaette My company originated the LMDB work (and LMDB).  I've kicked this back to the developer to investigate.

Your pull request does not seem acceptable as it adds tons of gentoo specific files.  Please limit the pull request to contain purely the code fix.

quanah:
@hhoffstaette sorry, ignore that last bit about the pull request, I clicked on your other reference by mistake. ;)

quanah:
@hhoffstaette Tested LMDB with cyrus-sasl git master this morning, it works correctly.

PR #557, opened 2019-03-14
Fix build with --disable-shared --enable-static

orivej:
1. Rename `SASL_STATIC_LIBS` to `SASL_STATIC_LIB` to free up `SASL_STATIC_LIBS` for
the list of additional libraries needed by static libsasl due to the plugins
that are included in libsasl only in the static build.
2. Add `LIB_DES` to `SASL_STATIC_LIBS` for the digestmd5 plugin.
3. Add `SASL_DB_LIB` to the list of libraries needed by libsasldb.

The second point fixes linking `utils/dbconverter-2` which fails with:
```
ld: ../lib/.libs/libsasl2.a(digestmd5.o): in function `init_3des':
lib/digestmd5.c:968: undefined reference to `DES_key_sched'
```

The third point fixes linking `sample/client` which fails with:
```
ld: ../lib/.libs/libsasl2.a(db_berkeley.o): in function `berkeleydb_open':
lib/db_berkeley.c:103: undefined reference to `db_create'
```

---

This was broken some time between 2.1.26 and fbfae51552f2972799a4cc8d1c25c5626a510ca6, which is the first commit I was able to build from source.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -126,10 +126,12 @@ AC_PROG_LIBTOOL
 target=$save_target
 
 if test "$enable_static" = yes; then
-	SASL_STATIC_LIBS=libsasl2.a
+	SASL_STATIC_LIB=libsasl2.a
 else
-	SASL_STATIC_LIBS=
+	SASL_STATIC_LIB=
 fi
+AC_SUBST(SASL_STATIC_LIB)
+SASL_STATIC_LIBS=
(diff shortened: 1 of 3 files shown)


quanah:

orivej:
OK.

hyc (approved):

PR #559, opened 2019-04-01
Use Openssl's EVP implementation of RC4 when available

simo5:
This patch let's us use the openssl provided RC4 via the sanctioned EVP interface.

It would be nice if we could completely drop custom implementations of cryptography, but for our uses it is enough that we do not have to use custom crypto by compiling it out.

I tested this patch via sample-server/sample-client forcing DIGEST-MD5 and minssf of 128, everything works as expected.

Please let me know if this is the right way to go to start replacing open coded cryptographic algorithms with the OpenSSL EVP interface, I would like to move DS and MD5 too eventually.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1103,13 +1103,12 @@ AC_ARG_WITH(configdir, [   --with-configdir=DIR    set the directory where confi
   configdir='${plugindir}:${sysconfdir}/sasl2')
 AC_SUBST(configdir)
 
-dnl look for rc4 libraries. we accept the CMU one or one from openSSL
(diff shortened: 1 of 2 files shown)


quanah:
The debug message change to simple server seems unrelated to this work?

simo5:
Yeah, I removed the change, didn't mean to change that

simo5:
Rebased on master and updated to avoid rc4 doublefree in openssl code path as well.

MehdiChinoune:
This change requires:
- Check that OpenSSL is enabled
- link digest to `lcrypto`

PR #562, opened 2019-04-16
Prevent double free of RC4 context

simo5:
free_rc4 can be called multiple times on the same context as reported in the mailing list.
This patch adds guards to NULL pointers after free is called.


nacho:
Looks good to me

PR #565, opened 2019-05-11
docsrc/exts/sphinxlocal: remove reference to docutils_version

dilyanpalauzov:
New sphinx does not contain docutils_version in sphinx.util.compat and manpage.py does
not use it, so just do not import it.


quanah:

quanah:
Requires DCO still
Needs rebasing on current master

bgermann:
The same change was integrated with ee592d7be6338bacac69657897deed3a715f199e. This can be closed.

PR #566, opened 2019-05-11
configure.ac: clarify --enable/--disable usage

dilyanpalauzov:
If a feature is enabled by default `./configure --help` should print '--disable-feature' and the called concludes from this output that the feature is enabled by default, even if the text after '--disable-feature' doesn't state this explicitly.

Likewise, if a feature is disabled by default `./configure --help` should print '--enable-feature' without stating, that the default is no.

* switch all calls of AC_ARG_ENABLE to use AC_HELP_STRING
* siwth DIGEST from AC_ARG_ENABLE to AC_ARG_WITH, because it accepts a parameter, which is unusual for AC_ARG_WITH
* exploit the fact that AC_ARG_ENABLE(x, text,,) sets $enable_x, when --enable-x is called, so don't set the variable explicitly
* reduce the calls of AC_ARG_RESULT
* make ./configure --disable-macos-framework do the right thing
* Remove unnecessary quoting, like "yes" → yes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -73,26 +73,22 @@ DIRS=""
 
 AC_ARG_ENABLE(cmulocal,
               [AC_HELP_STRING([--enable-cmulocal],
-                              [enable local mods for CMU [[no]]])],
-              [],
+                              [enable local mods for CMU])],,
               enable_cmulocal=no)
 
 AC_ARG_ENABLE(sample,
               [AC_HELP_STRING([--enable-sample],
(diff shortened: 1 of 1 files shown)


quanah:

It also needs to be updated since it now has conflicts.

quanah:
Requires DCO sign off still to be considered
Requires rebasing to work with current master

Neustradamus:
@dilyanpalauzov: Have you progressed on your PR?
Have you seen @quanah comments?

PR #575, opened 2019-07-02
Call gss_acquire_cred() before gss_init_sec_context() in gssapi_client_mech_step()

hrs-allbsd:
gss_init_sec_context() with GSS_C_NO_CREDENTIAL ignores a configured keytab by using gsskrb5_register_acceptor_identity().  This patch adds a gss_acquire_cred() call before gss_init_sec_context() to get a credential from the specified KRB5 keytab.

Note that gssapi_server_mech_step() already has the same code path and it works as expected while  the client version does not.  There is a case that a SASL client application uses a keytab.  However, the current code prevents it from working.  This change should have no harmful effect when not using gsskrb5_register_acceptor_identity().

This patch also includes a change to add a gsskrb5_register_acceptor_identity() call in gssapiv2_client_plug_init().   If gssapiv2_server_plug_init() is always called together, this is redundant.

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -1777,6 +1777,27 @@ static int gssapi_client_mech_step(void *conn_context,
 	    req_flags = req_flags |  GSS_C_DELEG_FLAG;
 	}
 
+	/* If caller didn't provide creds already */
+	if (client_creds == GSS_C_NO_CREDENTIAL) {
+	    GSS_LOCK_MUTEX_CTX(params->utils, text);
+	    maj_stat = gss_acquire_cred(&min_stat,
+					text->server_name,
+					GSS_C_INDEFINITE,
+					GSS_C_NO_OID_SET,
+					GSS_C_INITIATE,
+					&text->client_creds, 
+					NULL, 
+					NULL);
+	    GSS_UNLOCK_MUTEX_CTX(params->utils, text);
+
+	    if (GSS_ERROR(maj_stat)) {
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
With MIT Kerberos I put in krb5.conf
```
[realms]
X.NET = {
  kdc = mail.aegee.org
}
```
And then do
```
  kinit 1@X.NET
  Password for 1@X.NET: abc
```
(the password is abc).

With commit 238380260fe623212c0f21d63e763b7a849540d1
  Author: Hiroki Sato <hrs@allbsd.org>
  Date:   Tue Jul 2 15:32:29 2019 +0900

    Call gss_acquire_cred() before gss_init_sec_context() in
    gssapi_client_mech_step().  This fixes a bug that a keytab specified in
    gsskrb5_register_acceptor_identity() was not used on the client side.


reverted, calling `imtest -m GSSAPI -t "" mail.aegee.org`returns `Authenticated, Security strength factor: 256` and `klist` returns krbtgt/X.NET@X.NET, imap/mail.aegee.org@ and imap/mail.aegee.org@X.NET.

With the above commit applied (not reverted), the output after
```
  kdestroy
  kinit 1@X.NET
  imtest -m GSSAPI -t "" mail.aegee.org
```
is `Authentication failed. generic failure, Security strength factor: 256` and klist mentions only krbtgt/X.NET@X.NET.

`imtest -m GS2-KRB5 -t "" mail.aegee.org` does work.

PR #577, opened 2019-07-23
Prioritize SCRAM-SHA-256 over SCRAM-SHA-1

Ri0n:
scram-sha-256 seems to be the next one by strength. So +1 is fine.

### DIFF
--- a/plugins/scram.c
+++ b/plugins/scram.c
@@ -1845,7 +1845,7 @@ static sasl_server_plug_t scram_server_plugins[] =
 #ifdef HAVE_SHA256
     {
 	"SCRAM-SHA-256",		/* mech_name */
-	0,				/* max_ssf */
+	1,				/* max_ssf */
 	SASL_SEC_NOPLAINTEXT
 	| SASL_SEC_NOACTIVE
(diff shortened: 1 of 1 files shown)


ksmurchison:
Stating that SCRAM-SHA-256 has integrity protection (SSF=1) when it does not is not the correct way to prioritize the mechanisms.  We probably need to introduce a new way of prioritizing these mechanisms.

Neustradamus:
@ksmurchison: Thanks for your reply, it is very important to see for priority!

aamelnikov:
I think I just fixed this in commit 2088757976668324ae5ee1a87d912d7e2e2962dc. Can people double check that I didn't break anything?

aamelnikov:
Actually, I forgot to do a similar change in the client side of SCRAM.

Ri0n:
I'll check when ready.

aamelnikov:
I just pushed the client side change as well.

Ri0n:
seems like the mechanism is selected properly now.
Tested with Psi IM.

Neustradamus:
Thanks a lot @aamelnikov :)

PR #579, opened 2019-09-15
Ldap_tls_check_peer takes effect for no

tychn:
Setting the ldap_tls_check_peer parameter to no does not take effect. 
logical error, if  conf->tls_check_peer is 0, will not execute ldap_set_option, never set LDAP_OPT_X_TLS_REQUIRE_CERT to 0


quanah:
Definitely valid, this means people can't turn it off at all (and it defaults to on in the client side).

Issue #585, opened 2019-11-21
Feature idea: Make CyrusSASL store Kerberos tokens for delegation


aiobofh:
If one write a service with SASL and GSSAPI authentication, it would be awesome to have a way to forward/delegate that user authentication to other underlying services and subsystems.

I have a working (GSSProxy compatible) concept for this that allows for any service using CyrusSASL to make this a reality. A minor improvement to plugins/gssapi.c that make use of a few new key/value pairs for that service in the sasl2-configuration file for that service. E.g. where to store ccreds and a naming schema along with information on persistence for autonomous impersonation. 

Waiting for @aamelnikov on how to best contribute.

aiobofh:
@aamelnikov is the pull-request a good way to get the code in?

quanah:
merged

PR #586, opened 2019-11-28
plugins/gssapi: Implement credentials cache for delegation

aiobofh:
This feature enables CyrusSASL to forward/delegate GSSAPI Kerberos cached credentials for authenticating to other services on the server-side, as suggested in Issue #585. It also optionally implements service keytab loading with gss_acquire_creds_from() to enable privilege separation using e.g. GSSProxy on a POSIX server.

The feature is implemented on behalf of Vricon Systems AB

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -506,15 +506,19 @@ if test "$gssapi" != "no"; then
   load_from_default="no"
 
   AC_CHECK_HEADERS(gssapi.h, [have_gss_header=y])
-  AC_SEARCH_LIBS(gss_acquire_cred_from, [gssapi_krb5 gssapi], [have_load_from=t])
-  if test x"${have_gss_header}" != x -a x"${have_load_from}" != x; then
-     load_from_defaults="yes"
+  AC_CHECK_HEADERS(gssapi/gssapi_ext.h, [have_gss_ext_header=y])
+  AC_SEARCH_LIBS(gss_acquire_cred_from, [gssapi_krb5], [have_load_from=t])
+  AC_SEARCH_LIBS(gss_store_cred_into, [gssapi_krb5], [have_store_into=t])
+
+  if test x"${have_gss_ext_header}" != x -a x"${have_load_from}" != x; then
+     load_from_default="yes"
+  fi
+  if test x"${have_gss_ext_header}" != x -a x"${have_store_into}" != x; then
+     ccache_default="yes"
   fi
 
   if test "$gss_impl" = "mit"; then
      mutex_default="yes"
-     ccache_default="yes"
-     load_from_default="yes"
   elif test "$gss_impl" = "heimdal"; then
      AC_DEFINE(KRB5_HEIMDAL,[],[Using Heimdal]) 
   fi
--- a/configure.ac
+++ b/configure.ac
@@ -502,8 +502,19 @@ if test "$gssapi" != "no"; then
   AC_DEFINE(HAVE_GSSAPI,[],[Include GSSAPI/Kerberos 5 Support])
 
   mutex_default="no"
+  cc_default="no"
+  load_from_default="no"
+
+  AC_CHECK_HEADERS(gssapi.h, [have_gss_header=y])
+  AC_SEARCH_LIBS(gss_acquire_cred_from, [gssapi_krb5 gssapi], [have_load_from=t])
+  if test x"${have_gss_header}" != x -a x"${have_load_from}" != x; then
+     load_from_defaults="yes"
+  fi
+
   if test "$gss_impl" = "mit"; then
      mutex_default="yes"
+     cc_default="yes"
+     load_from_default="yes"
   elif test "$gss_impl" = "heimdal"; then
      AC_DEFINE(KRB5_HEIMDAL,[],[Using Heimdal]) 
   fi
@@ -516,6 +527,26 @@ if test "$gssapi" != "no"; then
      AC_DEFINE(GSS_USE_MUTEXES, [], [should we mutex-wrap calls into the GSS library?])
   fi
   AC_MSG_RESULT($use_gss_mutexes)
+
+  AC_MSG_CHECKING(to use GSS delgation with ccpath)
+  AC_ARG_ENABLE(gss_cc, [  --enable-gss-cc         allow delegation of cached crdentials],
+                use_gss_cc=$enableval,
+                use_gss_cc=$cc_default)
+
+  if test $use_gss_cc = "yes"; then
+     AC_DEFINE(GSS_USE_CC, [], [should we allow storage of cached credentials?])
+  fi
+  AC_MSG_RESULT($use_gss_cc)
+
+  AC_MSG_CHECKING(to use GSS service keytab loading with gss_acquire_cred_from)
+  AC_ARG_ENABLE(gss_load_from, [  --enable-gss-load-from  allow GSSProxy friendly keytab load],
+                use_gss_load_from=$enableval,
+                use_gss_load_from=$load_from_default)
+
+  if test $use_gss_load_from = "yes"; then
+     AC_DEFINE(GSS_USE_LOAD_FROM, [], [should we use gss_acquire_cred_from?])
+  fi
+  AC_MSG_RESULT($use_gss_load_from)
 fi
 
 SASL2_CRYPT_CHK
@@ -1218,6 +1249,30 @@ if test $sasl_cv_getnameinfo = yes; then
 fi
 AC_SUBST(GETNAMEINFOOBJS)
 AC_SUBST(LTGETNAMEINFOOBJS)
+
+dnl check for getpwnam
+sasl_cv_getpwnam=no
+AC_CHECK_FUNC(getpwnam, [AC_DEFINE(HAVE_GETPWNAM,[],
+	[do we have getpwnam()?])], [sasl_cv_getpwnam=yes])
+AC_SUBST(GETPWNAM)
+
+dnl check for getuid
+sasl_cv_getuid=no
+AC_CHECK_FUNC(getuid, [AC_DEFINE(HAVE_GETUID,[],
+	[do we have getuid()?])], [sasl_cv_getuid=yes])
+AC_SUBST(GETUID)
+
+dnl check for getlogin_r
+sasl_cv_getlogin_r=no
+AC_CHECK_FUNC(getlogin_r, [AC_DEFINE(HAVE_GETLOGIN_R,[],
+	[do we have getlogin_r()?])], [sasl_cv_getlogin_r=yes])
+AC_SUBST(GETLOGIN_R)
+
+dnl check for getpid
+sasl_cv_getpid=no
+AC_CHECK_FUNC(getpid, [AC_DEFINE(HAVE_GETPID,[],
+	[do we have getpid()?])], [sasl_cv_getpid=yes])
+AC_SUBST(GETPID)
 
 LTLIBOBJS=`echo "$LIB@&t@OBJS" | sed 's,\.[[^.]]* ,.lo ,g;s,\.[[^.]]*$,.lo,'`
 AM_CONDITIONAL(BUILD_LIBOBJ, test "$LTLIBOBJS" != "")
--- a/configure.ac
+++ b/configure.ac
@@ -502,7 +502,7 @@ if test "$gssapi" != "no"; then
   AC_DEFINE(HAVE_GSSAPI,[],[Include GSSAPI/Kerberos 5 Support])
 
   mutex_default="no"
-  cc_default="no"
+  ccache_default="no"
   load_from_default="no"
 
   AC_CHECK_HEADERS(gssapi.h, [have_gss_header=y])
@@ -513,7 +513,7 @@ if test "$gssapi" != "no"; then
 
   if test "$gss_impl" = "mit"; then
      mutex_default="yes"
-     cc_default="yes"
+     ccache_default="yes"
      load_from_default="yes"
   elif test "$gss_impl" = "heimdal"; then
      AC_DEFINE(KRB5_HEIMDAL,[],[Using Heimdal]) 
@@ -528,15 +528,15 @@ if test "$gssapi" != "no"; then
   fi
   AC_MSG_RESULT($use_gss_mutexes)
 
-  AC_MSG_CHECKING(to use GSS delgation with ccpath)
-  AC_ARG_ENABLE(gss_cc, [  --enable-gss-cc         allow delegation of cached crdentials],
-                use_gss_cc=$enableval,
-                use_gss_cc=$cc_default)
+  AC_MSG_CHECKING(to use GSS delgation with ccache store)
+  AC_ARG_ENABLE(gss_ccache, [  --enable-gss-ccache       allow delegation of cached crdentials],
+                use_gss_ccache=$enableval,
+                use_gss_ccache=$ccache_default)
 
-  if test $use_gss_cc = "yes"; then
-     AC_DEFINE(GSS_USE_CC, [], [should we allow storage of cached credentials?])
+  if test $use_gss_ccache = "yes"; then
+     AC_DEFINE(GSS_USE_CCACHE_STORE, [], [should we allow storage of cached credentials?])
   fi
-  AC_MSG_RESULT($use_gss_cc)
+  AC_MSG_RESULT($use_gss_ccache)
 
   AC_MSG_CHECKING(to use GSS service keytab loading with gss_acquire_cred_from)
   AC_ARG_ENABLE(gss_load_from, [  --enable-gss-load-from  allow GSSProxy friendly keytab load],
--- a/docsrc/sasl/options.rst
+++ b/docsrc/sasl/options.rst
@@ -114,7 +114,7 @@ GSSAPI
 
    If this option is omitted, credentials delegation will not be enabled.
 
-   See MIT Kerberos documentation on more informatino about credentials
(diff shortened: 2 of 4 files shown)


aiobofh:
Hm, some Ruby problems in Travis.

plugins/gssapi.c
```diff
@@ -811,6 +829,229 @@ gssapi_server_mech_new(void *glob_context,
...
+ malloc_failed:
+ sanity_check_failed:
+ parsing_failed:
+  if (NULL != ccname) {
+    free(ccname);
+    ccname = NULL;
+  }
+ noccache:
+  if (NULL == ccname) {
+    ccname = strdup(getenv("KRB5CCNAME"));
```
simo5: Nothing guarantees a KRB5CCNAME is always set, libkrb5 does not depend on it, as it can use the default ccache defined in krb5.conf that can use substitutions too.
You should not fail if KRB5CCNAMe is empty and  just let the default ccache be used I think.

plugins/gssapi.c
```diff
@@ -811,6 +829,229 @@ gssapi_server_mech_new(void *glob_context,
...
+
+  strncpy(d, haystack, pre_needle_len);
+  d[pre_needle_len] = 0;
+  strcat(d, subst);
+  strcat(d, &haystack[pre_needle_len + needle_len]);
+
+  return d;
+}
+
+static char* create_krb5_ccache(context_t* text, const char* username, const char* ccache) {
```
simo5: You are not creating the ccache here, simply using a tempate to provide a ccache name, suggest changing name to reflect that and avoid confusion.
Something like resolve_ccache_name()

plugins/gssapi.c
```diff
@@ -1093,6 +1371,23 @@ gssapi_server_mech_authneg(context_t *text,
 	goto cleanup;
     }
 
+#ifdef GSS_USE_CC
+    /* Get ccache from sasl-config file if located elsewhere than /tmp */
```
simo5: I do not understand the comment, in many systems ccache has been moved to different types (KEYRING, KCM, etc.. for quite some time, do you mean "if not the default ccache" ? If so I would phrase it that way, mentioning /tmp is confusing (even on systems where a FILE ccache is used the /tmp location has been dscouraged for a long time already)

docsrc/sasl/options.rst
```diff
@@ -96,6 +96,30 @@ GSSAPI
 
    Default: /etc/krb5.keytab (system dependant)
 
+.. option:: ccache [<path>]
```
simo5: The term cache is exceedingly vague, I suggest something like client_ccache or delegated_ccache_store or similar ...

configure.ac
```diff
@@ -564,6 +575,26 @@ if test "$gssapi" != "no"; then
      AC_DEFINE(GSS_USE_MUTEXES, [], [should we mutex-wrap calls into the GSS library?])
   fi
   AC_MSG_RESULT($use_gss_mutexes)
+
+  AC_MSG_CHECKING(to use GSS delgation with ccpath)
+  AC_ARG_ENABLE(gss_cc, [  --enable-gss-cc         allow delegation of cached crdentials],
+                use_gss_cc=$enableval,
+                use_gss_cc=$cc_default)
+
+  if test $use_gss_cc = "yes"; then
```
simo5: GSS_USE_CC is very vague, I would use a more descriptive define like GSS_STORE_DELEGATED_CREDS

plugins/gssapi.c
```diff
@@ -811,6 +829,229 @@ gssapi_server_mech_new(void *glob_context,
...
+static char *store_client_credentials(context_t* text, gss_cred_id_t* client_credentials, const char* username, const char* ccache) {
+  gss_key_value_element_desc element;
+  gss_key_value_set_desc store;
+  char* ccname = NULL;
+  OM_uint32 maj_stat;
+  OM_uint32 min_stat;
+
+  if (NULL == (ccname = create_krb5_ccache(text, username, ccache))) {
+    text->utils->seterror(text->utils->conn, SASL_LOG_ERR,
+			  "GSSAPI error: unable to generate krb5cc filename");
```
simo5: filename -> cache name

plugins/gssapi.c
```diff
@@ -877,15 +1118,48 @@ gssapi_server_mech_authneg(context_t *text,
...
-					text->server_name,
-					GSS_C_INDEFINITE, 
-					GSS_C_NO_OID_SET,
-					GSS_C_ACCEPT,
-					&text->server_creds, 
-					NULL, 
-					NULL);
+#ifdef GSS_USE_LOAD_FROM
+            if (keytab != NULL) { /* Force cached credentials to use GSS Proxy compliant API of GSS */
+                gss_key_value_element_desc ccache_element = {.key = "ccache", .value = NULL};
```
simo5: Why are you setting ccache to NULL here?
I know NULL values currently work ( I wrote this API) but they are an undefined behavior and should not be used.
It may crash the program in the future.

aiobofh: Right... Hm. I have mostly (only) seen this kind of approach in other project just about everywhere. I just suspected it would do some kind of fallback. Any good ideas on how to make it correct? I don't think it should be used at all this early (since it's basically overridden once a user has authenticated by the other code).

Also, I know it's a bit awekward to load the service creds like this (when a user connects), i know. But for now it will do. - Could be up for a move to program initialization at some point in the future.

plugins/gssapi.c
aiobofh: Heh :) Yeah, be gone with it :+1:

plugins/gssapi.c
aiobofh: I guess; what should be better is if this resolve-function wont succeed, the calling code should make the decision to resolve using the krb5.conf settings/default in that case somehow. Do you have a nice little suggestion on how to get hold of it if it's not in the ENV? I was kind-of hoping automagic fallback to the host default config.

plugins/gssapi.c
aiobofh: I guess the "ccache" key/value (or rather KRB5_CS_CCACHE_URN) could be omitted altogether. *thinking*

aiobofh: There was also a weird assignment that got immediately overridden there... Cleaning this up a bit. Good eye!

simo5: Just drop setting the ccache value if you do not want to set it to a specific value, the NULL value, while currently accepted by the MIT implementation is an invalid input.

plugins/gssapi.c
simo5: If your plan is to use the defaults, then do not do anything here.
gssapi/krb5 will always use KRB5CCNAME and fallback to krb5.conf profile options on its own already.

aiobofh: Great, thanks. :+1:

plugins/gssapi.c
aiobofh: Yeah, i think the copy'n'paste monster took a dump in there :) Making it so.

Just need to run a regression run before pushing.

aiobofh: Sorry for the delay... Got stuck with it for a while. Something seems borked on the client side from either 7a95382c68e7187fe7407b2a94036d9ca3246e34 or 238380260fe623212c0f21d63e763b7a849540d1 - At least I cannot test like I did before (I initially based my work on the latest release, now it's branched from master), by just initializing a default principal / ticket cache with kinit and then start the sample-client. Have to use the last released one, client side, where it works, for testing. Took me a while to figure out what bailed on me tho. @simo5 , see any reason why? It does not feel like an odd use-case. Just using the default available token to authenticate using GSSAPI :) It now exits since gss_acquire_cred() fails, or maybe as a consequence for gsskrb5_register_acceptor_identity() failure. (This is however unrelated to this pull-request besides the fact that it seem to complicate use of GSSAPI for next release - or more likely; I'm misunderstanding something).

simo5: In general gsskrb5_register_acceptor_identity() should be avoided as it is basically an irreversible operation, and it is not needed with modern APIs like load_from/store_into you are trying to use here.

Also why those patches limit the path to 1024 chars in 2019? We have PATH_MAX available if really needed (which is defined as 4096 on linux for example), and in general allocating on the heap would be preferred, no point in wasting 1k of stack like that ... but I digress.

Please do make gss_acquire_cred_from mutually exclusive with gssapi_register_acceptor_identity() with the former being the preferred feature to use and register identity only as an optional fallback.

aiobofh: Right... Just to be clear.... I did add gss_acquire_cred_from server-side. The problem I described above with the client side is a contribution by someone else, back in July, but it makes it (for me) "un-possible" to test my server-side changes. With what's on master/my fork right now :)

aiobofh: @simo5 yay! #591 will fix the issue with not reading the default creds. However just getting _this_ pull request into master will make the delegation feature-incomplete. Both are needed.

plugins/gssapi.c
aiobofh: @simo5  I'm trying to make a sensible fall-back flow according to https://web.mit.edu/kerberos/krb5-1.12/doc/basic/ccache_def.html#default-ccache-name but I wanted to take the chance to ask you if you know of any API in the GSSAPI-part of KRB5 that would give me this automagically, have not found anything relevant.

docsrc/sasl/options.rst
aiobofh: Settled for ccache_store instead. It's similar enough to other systems I've seen.

plugins/gssapi.c
aiobofh: ... Also... I'm thinking. The gss_store_cred_into (which I use) is relying on the configurable stuff a bit more. Maybe if there is a gss_store_cred() call as fallback and reduced API capabilities it should handle the fall-back flow, right?

plugins/gssapi.c
```diff
@@ -1093,6 +1398,31 @@ gssapi_server_mech_authneg(context_t *text,
...
+        const char* ccache = NULL;
+        const sasl_utils_t *utils = text->utils;
+        unsigned int rl;
+        utils->getopt(utils->getopt_context, "GSSAPI", "ccache_store", (const char**)&ccache, &rl);
+        if (NULL == ccache) {
+            text->ccname = NULL;
+        }
+        else {
+            text->ccname = store_client_credentials(text, &(text->client_creds),
+                                                    name_token.value, ccache);
```
simo5: There is no guarantee that the value of name_token is zero terminated (in fact I know for a fact there are cases where it is not).
This is a bug throughout this function apparently and should be independently fixed (separate commit?)

But given you later want to use the gss name to resolve users, you may want to change the code to pass down name_token  (maybe change to use a better name?) whole here so you can later call gss_localname() on it.

plugins/gssapi.c
```diff
@@ -1706,7 +2036,7 @@ static int gssapi_client_mech_step(void *conn_context,
 		
 		return SASL_INTERACT;
 	    }
-	}
+        }
```
simo5: indentation change?

plugins/gssapi.c
```diff
@@ -894,7 +894,32 @@ static char* resolve_ccache_store(context_t* text, const char* username, const c
         case 'u': {
 #ifdef HAVE_GETPWNAM
             char uid_str[100];
-            struct passwd* pwd = getpwnam(username);
+            struct passwd* pwd;
+            size_t username_len;
+            char* username_wo_realm;
+            const char* at_sign_at = strchr(username, '@');
+            /* Strip @REALM part, if available, since password database most likely wont contain it */
```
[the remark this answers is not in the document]
aiobofh: @simo5 good idea! I'll make it so.

plugins/gssapi.c
```diff
@@ -1261,9 +1286,8 @@ gssapi_server_mech_authneg(context_t *text,
 	text->qop = LAYER_NONE | LAYER_INTEGRITY | LAYER_CONFIDENTIALITY;
     }
 #ifdef GSS_USE_CCACHE_STORE
-    if (text->ccname) {
-        params->props.security_flags |= SASL_SEC_PASS_CREDENTIALS; /* Force delegation flag */
-    }
+    /* TODO: This should maybe depend on "pass_credentials" (server-side), or Kerberos token attributes */
```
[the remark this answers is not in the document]
aiobofh: It's the old commit with the if-statement. It's gone.

configure.ac
```diff
@@ -1267,6 +1302,30 @@ fi
...
+	[do we have getuid()?])], [sasl_cv_getuid=yes])
+AC_SUBST(GETUID)
+
+dnl check for getlogin_r
+sasl_cv_getlogin_r=no
+AC_CHECK_FUNC(getlogin_r, [AC_DEFINE(HAVE_GETLOGIN_R,[],
+	[do we have getlogin_r()?])], [sasl_cv_getlogin_r=yes])
+AC_SUBST(GETLOGIN_R)
+
+dnl check for getpid
```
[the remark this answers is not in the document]
aiobofh: Hmm.. Maybe that's overkill... Don't know actually.

plugins/gssapi.c
aiobofh: Right.. Yeah, it's just passed through two levels, so maybe that's a more logical design. I'll take a looksie. It is actually used, in the lower functions where a non-null-terminated string would be bad.

@simo5 a shame it's like that. Brings unnecessary boilerplate into the implementations. Feels like a bad design choice in whatever library implementation. Anyhow; I totally agree. I'll fix.

plugins/gssapi.c
aiobofh: Hmm.. Yeah, I saw that one at a glance when making the commit, but its a false positive, i think. 300 lines away... I'll double check.

plugins/gssapi.c
aiobofh: @simo5 I don't feel i have (paid) time to refactor the whole thing (but I will make an effort on the new code I contributed), there are several strdup() and other things operating on only the .value element of the gss_buffer_desc elsewhere in gssapi_server_mech_authneg(). I could write some issues on various code-lines I've seen with similar use as I replicated tho.

quanah:
Just curious, have you tested this with Heimdal?

aiobofh:
@quanah that's a no.

configure.ac should enable features as specific functions exist in the API. So, if they're not in Heimdal it won't be enabled, or if its implemented a different way than MIT - it will probably fail.

plugins/gssapi.c
aiobofh: I really had to configure my Emacs in a weird way to get it to like the code and make indentation as similar as possible to what's already there... There was a TAB there before, somewhere along the way it got replaced by spaces. Fixed it so it's like before, anyway.

aiobofh:
Thank you very much, @simo5, for reviewing!

quanah:
I would like to see this expanded to cover Heimdal as well.  It should be doable using the gss_acquire_cred function.  You can see an example of this in Samba's code base: https://github.com/samba-team/samba/blob/master/lib/krb5_wrap/gss_samba.c.

aiobofh:
@quanah i don't have funding to cover Heimdal right now, nor the time. Better with something than nothing.

simo5:
@quanah recent Heimdal does have gss_acquire_cred_from()/gss_store_cred_into() support is that sufficient to you ?

simo5:
Reference: https://github.com/heimdal/heimdal/pull/492

aiobofh:
The configure.ac might need some changes tho. I think I did only explicitly search for those API functions when MIT Kerberos was detected.

quanah:
Those functions are not in any Heimdal release nor is there a timeline on when they will be available (I asked the project head).  It might be worthwhile to test against Heimdal master I suppose, and adjust configure.ac to detect it for when they become available at some point in the future, and additionally that the code works as expected.

simo5:
Sounds reasonable

aiobofh:
@simo5 and @quanah ok - i have my forked repo in sync with the main repo now... maybe i did a silly kind of rebase when i come to think of it... anyhow. I hope it helps you with your work.

simo5:
@aiobofh something is wrong, you have 42 commits and 103 file changes, that ... doesn't help ...
Any chance you can rebase only your work on top of master? Seem like you pulled in a truckload of merges, that's probably because you did a git pull from master in your repo ... never do that, my suggestion is to rewind back (look at the reflog) and then do a git pull --rebase

aiobofh:
@simo5 hahaha :) yeah. Shite. Seeing this... It turned out like a huge turd. Never do stuff tired/stressed. *embarrassed* 

I'll see if i can get this under control... I'm not a git-meister for this kind of things tho :/ I'll try to figure it out.

I rewinded 29 commits on my local branch, i guess i need to do a push --force before trying any git pull --rebase. otherwise I'll just end-up in the same situation again, right?

simo5:
You need to push --force only once you are done locally, what's in your remote (the on on github) repo doesn't influence what you do on your local machine at all, it is just a copy.

aiobofh:
@simo5 sorry for the git-support questions :(

Better now?

aiobofh:
@simo5 I'm sorry - I have not had a minute over last month to even try out your test-framework on this feature. :( I HOPE to get time still. But this super-plus-bad-flu thing is disrupting.

quanah:

aiobofh:
@quanah Done!  (i think :D )

megahallon:
What is the status of this one, could it be merged?

quanah:
we are in the process of reviewing all open pull requests.  We haven't reached this one yet.

PR #593, opened 2020-02-24
Revert the change to acquire init credentials

iboukris:
Hello, I think the change to call gss_acquire_cred() in the client code, is not correct.

Commit 238380260fe623212c0f21d63e763b7a849540d1 broke GSSAPI kerberos authentication, and commit 74faca7400f414784b5e2e136668e6f4ef0d6b96 "fixed" it by ignoring the error (which in my opinion will always occur, as below).

First, calling gss_acquire_cred() to acquire initiator client credentials and passing 'text->server_name' as client desired_name would never make sense, unless the only use case is when the client and server are the same principal, i think this can never work.

Secondly, gsskrb5_register_acceptor_identity() is for acceptor keytab, it is not for initiator, if it makes any impact then it may be a bug.
Instead, you can just use 'default_client_keytab_name' which is supported on modern MIT and and Heimdal, and then you don't need this change.
If you really want to control which keytab to use from code, you can use gss_acquire_cred_from() from the "credential store" API, which is also supported on both MIT and Heimdal implementations.

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -1777,17 +1777,7 @@ static int gssapi_client_mech_step(void *conn_context,
 	    req_flags = req_flags |  GSS_C_DELEG_FLAG;
 	}
 
-	/*
-	 * If caller didn't provide creds already.
-	 *
-	 * In the case of Kerberos, a client typically wants to use
-	 * a credential in either a keytab file or the credentials cache
-	 * of the current process context.  This code path will try to
-	 * find a credential in the specified keytab file,  then the
-	 * credentials cache.  The keytab file can be specified by
-	 * "keytab" option, and it is configured by using
-	 * gsskrb5_register_acceptor_identity() API when available.
-	 */
+	/* If caller didn't provide creds already */
 	if (client_creds == GSS_C_NO_CREDENTIAL) {
 	    GSS_LOCK_MUTEX_CTX(params->utils, text);
 	    maj_stat = gss_acquire_cred(&min_stat,
@@ -1800,16 +1790,14 @@ static int gssapi_client_mech_step(void *conn_context,
 					NULL);
 	    GSS_UNLOCK_MUTEX_CTX(params->utils, text);
 
-	    /*
-	     * Ignore the error intentionally.  The credential was
-	     * not found in the specified keytab file.
-	     */
-	    if (GSS_ERROR(maj_stat) == 0) {
-		client_creds = text->client_creds;
+	    if (GSS_ERROR(maj_stat)) {
+		sasl_gss_seterror(text->utils, maj_stat, min_stat);
+		sasl_gss_free_context_contents(text);
+		return SASL_FAIL;
 	    }
+	    client_creds = text->client_creds;
 	}
 
-	/* Try the credentials cache. */
 	GSS_LOCK_MUTEX_CTX(params->utils, text);
 	maj_stat = gss_init_sec_context(&min_stat,
 					client_creds, /* GSS_C_NO_CREDENTIAL */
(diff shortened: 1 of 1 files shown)


simo5:
Please coordinate with PR #586 to create the right fix.
But I do agree on these patches not being optimal, I think I hinted about that already in #586

iboukris:
Hi Simo, thanks I'll look in more details, but I think the bug i clear and this is a simple fix for something that could have never worked, so I think it may still make sense to revert first.

iboukris:
@simo5 looking at #586 it also adds features more code changes, so I think it would be better to first revert the none working code and then add on a clean branch.

simo5:
I tend to agree, I am fine if #586 is rebased on top of this one, just saying it needs to be looked at so pinging @aiobofh as well

aiobofh:
@simo5 and @iboukris hi guys! so.. what's up?

@simo5 What was the hint in #586 - did I forget to fix something? :cry: 

As I mentioned on the mailing list some time ago... Either none or both patches from @hrs-allbsd are required for the cached credentials to work, I think. So. I think #586 should just be to slab on to master, given that #2383802 is reverted or #591 is applied (before merge), depending on the satisfaction of #591, that is.

hrs-allbsd:
I admit calling a gss_acquire_cred() for initiator is a hack which works with Heimdal, and gss_acquire_cred_from() is a far reasonable way to implement it as pointed out.  One problem is that it is a relatively new API.

OK, so I will create another patch which can support platforms with and without gss_acquire_cred_from().  I still believe the committed change does not break typical configurations, but I do not mind reverting my changes completely.

simo5:
@aiobofh nothing wrong in #586, I just think it should be rebased on top of this PR an hopefully that is not hard.

@hrs-allbsd I think it would be more productive to merge this PR, and then propose a new one on top of #586 once that one is merge too. Please mention me when you propose your new PR as I have intimate knowledge of GSSAPI and can provide reviews as I did on #586

iboukris:
>  I still believe the committed change does not break typical configurations,

@hrs-allbsd,  I beg to differ.   Even if this change somehow works for you (which I really can't see how, even on Heimdal it is probably a bug), it would still be wrong to use a keytab and then fallback to ccache. A keytab is not a cache, if you force using it, you'd trigger AS-REQ followed by TGS-REQ, instead of just using a ticket from cache. I suggest to use gss_acquire_cred_from() with "client_keytab" and the krb5 library will do all the magic for you.

As the code is now, the call to gss_acquire_cred() *always* fails with MIT, causing spurious trace errors.

iboukris:
And it should fail on Heimdal too, unless there is a bug causing 'desired_name' to be ignored, as above.

aiobofh:
@simo5 sure! Yeah, it has been staying on the branch for quite some time :D I can give it a looksie later this week - I'm up to my neck with other work and projects.

hrs-allbsd:
My use case is that the server and the client share the same service principal for a automated task.  It is not a normal case because there is no ccache on the client side.

simo5:
Sorry @hrs-allbsd but your case is nothing special just a matter of using the same principal for two tasks. Moreover a client without ccache is simply impossible as you always need a ccache to store at least a direct AS request against the target service from the client.
I think we should revert because the code is not applicable to the generic case and it is not the correct way to resolve your issue either.

iboukris:
@hrs-allbsd   I don't think this esoteric use case would justify these changes, and you should be able to get it working with 'default_client_keytab_name', config or variable. If code changes are really needed, please avoid unnecessary error path, and make sure it works with both Heimdal and MIT.

quanah:
Explicitly which commits should be reverted?  Thanks!

simo5:
@quanah if you apply this PR, the right commits will be reverted

iboukris:
Yeah, the three commit from this PR, thanks!

quanah:
Thanks, updated in master and the cyrus-sasl-2.1 branches.

PR #598, opened 2020-03-18
Add basic test infrastructure

### LINKED ISSUE #595: Test infrastucture and github CI workflow?
I would like to start a conversation with upstream about adding some basic test infrastructure to test that SASL keeps working in common scenarios when PRs are proposed and using github CI workflows (for example running tests in Travis).

I am willing to provide some work towards PRs to implement this with some basic scenarios I may care about. Would there be interest in this?

simo5:
This runs in make check so hopefully it will work as CI via travis

First test is for SASL/GSSAPI

Fixes #595

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ m4/ltsugar.m4
 m4/ltversion.m4
 m4/lt~obsolete.m4
 .sphinx-build.stamp
+.tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,18 @@
 language: c
-sudo: required
-dist: trusty
 group: edge
 
 os:
   - linux
   - osx
 compiler: gcc
+
+addons:
+  apt:
+    packages:
+    - krb5-kdc
+    - krb5-admin-server
+    - libsocket-wrapper
+    - libnss-wrapper
 
 # For OS X, but shouldn't cause any problems on Linux
 cache:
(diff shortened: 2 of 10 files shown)


simo5:
@quanah can you take a look at this ?

iboukris:
Cool!

quanah:
@simo5 Yes, will look in more detail shortly. :)

simo5:
@quanah if this works, we can ask #586 to add a test to validate it works, and I can start adding more tests for other parts of SASL as well

aiobofh:
@simo5 awsome work! I'll try to find some time somewhere to write some tests on credential delegation and maybe loading of specified store. Can't give a deadline - Work is taking a bit more time now when the world is what it is. But hopefully soon...

quanah:
@simo5 For the 2.1 release branch, is this OK to add as-is, or do we need to tweak it?

simo5:
@quanah With the caveat that tests are not run on macosx (because I do not have one and travis is too slow to do iterative experiments) I do not see a problem having them in the release.
It does add a dependency in the make check step though.
If that is a problem we can add a make tests target instead so that ppl that do not want to have that dependency in the way can avoid it.

simo5:
@quanah aaand I just realized some files have this as the header:
`/* TBD, add (C) */`
I can follow up with a PR that fixes that, what License is acceptable? Should I add CMUs one ?

quanah:
@simo5 I would ask on the cyrus devel list for the license question

PR #599, opened 2020-03-19
Use proper length for Fully Qualified Domain Names

### LINKED ISSUE #583: Determination of own FQDN fails on GSSAPI server when FQDN is 64 characters or longer
Currently the `get_fqhostname` function assumes that the FQDN can be no longer than `MAXHOSTNAMELEN`. This is an incorrect assumption -- that constant (typically 64) only applies to the max _unqualified_ hostname. In fact, it's legal to have a FQDN which is 253 bytes long.

This results in truncation of the determined hostname.

In addition, the truncation is performed using `strncpy()` from the result if the DNS lookup into the result buffer, which does _not_ null terminate the result in the case that the result buffer doesn't have enough space. Then in `_sasl_conn_init` this non-null-terminated buffer is used as an argument to `strdup` meaning it ends up appending some garbage bytes from the stack following the buffer (until it happens to hit a null byte). The resulting error message then looks something like `No key table entry found matching impala/multidistro-191030-070716-e7r-dw3-coordinator0.multidis.xcu2-8y8xwˆyÿ@)` (note the junk at the end).

As a workaround, users can pass in an explicit FQDN.

simo5:
Per RFC 1035 the maximum FQDN is 253 chars, and not MAXHOSTNAMELEN which
varies between 64 and 1024 bytes depending on the system used.

Define and use a new variable.
Also fix get_fqhostname() to insure that the name returned is zero
terminated.

Fixes #583

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1442,11 +1442,6 @@ struct sockaddr_storage {
 #define NI_NUMERICSERV  8
 #endif
 
-/* Defined in RFC 1035. max strlen is only 253 due to length bytes. */
-#ifndef MAXHOSTNAMELEN
-#define        MAXHOSTNAMELEN  255
-#endif
-
 #ifndef HAVE_SYSEXITS_H
 #include "exits.h"
 #else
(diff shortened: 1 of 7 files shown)


simo5:
@quanah can we get this one in ?

PR #601, opened 2020-03-20
Add GSSAPI Channel Binding support

### LINKED ISSUE #600: GSSAPI module need ssupport for Channel Bindings
This is needed by openldap as new Windows Active Directory servers are now starting to require channel bindings when performing SASL/GSSAPI or SASL/GSS-SPENGO authentication in a TLS channel.

simo5:
This works and comes with both a positive and a negative test.

Fixes #600

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -830,7 +830,9 @@ gssapi_server_mech_authneg(context_t *text,
     gss_buffer_desc name_without_realm;
     gss_name_t client_name_MN = NULL, without = NULL;
     gss_OID mech_type;
-	
+    gss_channel_bindings_t bindings = GSS_C_NO_CHANNEL_BINDINGS;
+    struct gss_channel_bindings_struct cb = {0};
+
     input_token = &real_input_token;
     output_token = &real_output_token;
     output_token->value = NULL; output_token->length = 0;
@@ -902,6 +904,12 @@ gssapi_server_mech_authneg(context_t *text,
 	real_input_token.length = clientinlen;
     }
 
+    if (params->cbinding != NULL) {
+        cb.application_data.length = params->cbinding->len;
+        cb.application_data.value = params->cbinding->data;
+        bindings = &cb;
+    }
+
 
     GSS_LOCK_MUTEX_CTX(params->utils, text);
     maj_stat =
@@ -909,7 +917,7 @@ gssapi_server_mech_authneg(context_t *text,
 			       &(text->gss_ctx),
 			       server_creds,
 			       input_token,
-			       GSS_C_NO_CHANNEL_BINDINGS,
+			       bindings,
 			       &text->client_name,
 			       &mech_type,
 			       output_token,
@@ -1505,7 +1513,8 @@ static sasl_server_plug_t gssapi_server_plugins[] =
 	| SASL_SEC_PASS_CREDENTIALS,
 	SASL_FEAT_WANT_CLIENT_FIRST
 	| SASL_FEAT_ALLOWS_PROXY
-	| SASL_FEAT_DONTUSE_USERPASSWD,	/* features */
+	| SASL_FEAT_DONTUSE_USERPASSWD
+	| SASL_FEAT_CHANNEL_BINDING,	/* features */
 	NULL,				/* glob_context */
 	&gssapi_server_mech_new,	/* mech_new */
 	&gssapi_server_mech_step,	/* mech_step */
@@ -1529,6 +1538,7 @@ static sasl_server_plug_t gssapi_server_plugins[] =
 	SASL_FEAT_WANT_CLIENT_FIRST
 	| SASL_FEAT_ALLOWS_PROXY
 	| SASL_FEAT_DONTUSE_USERPASSWD
+	| SASL_FEAT_CHANNEL_BINDING
 	| SASL_FEAT_SUPPORTS_HTTP,	/* features */
 	&gss_spnego_oid,		/* glob_context */
 	&gssapi_server_mech_new,	/* mech_new */
@@ -1662,6 +1672,8 @@ static int gssapi_client_mech_step(void *conn_context,
     input_token->value = NULL; 
     input_token->length = 0;
     gss_cred_id_t client_creds = (gss_cred_id_t)params->gss_creds;
+    gss_channel_bindings_t bindings = GSS_C_NO_CHANNEL_BINDINGS;
+    struct gss_channel_bindings_struct cb = {0};
 
     if (clientout)
         *clientout = NULL;
@@ -1777,6 +1789,12 @@ static int gssapi_client_mech_step(void *conn_context,
 	    req_flags = req_flags |  GSS_C_DELEG_FLAG;
 	}
 
+        if (params->cbinding != NULL) {
+            cb.application_data.length = params->cbinding->len;
+            cb.application_data.value = params->cbinding->data;
+            bindings = &cb;
+        }
+
 	GSS_LOCK_MUTEX_CTX(params->utils, text);
 	maj_stat = gss_init_sec_context(&min_stat,
 					client_creds, /* GSS_C_NO_CREDENTIAL */
@@ -1785,7 +1803,7 @@ static int gssapi_client_mech_step(void *conn_context,
 					text->mech_type,
 					req_flags,
 					0,
-					GSS_C_NO_CHANNEL_BINDINGS,
+					bindings,
 					input_token,
 					NULL,
 					output_token,
@@ -2190,7 +2208,8 @@ static sasl_client_plug_t gssapi_client_plugins[] =
 	| SASL_SEC_PASS_CREDENTIALS,    /* security_flags */
 	SASL_FEAT_NEEDSERVERFQDN
 	| SASL_FEAT_WANT_CLIENT_FIRST
-	| SASL_FEAT_ALLOWS_PROXY,	/* features */
+	| SASL_FEAT_ALLOWS_PROXY
+	| SASL_FEAT_CHANNEL_BINDING,	/* features */
 	gssapi_required_prompts,	/* required_prompts */
 	GSS_C_NO_OID,			/* glob_context */
 	&gssapi_client_mech_new,	/* mech_new */
@@ -2213,6 +2232,7 @@ static sasl_client_plug_t gssapi_client_plugins[] =
 	SASL_FEAT_NEEDSERVERFQDN
 	| SASL_FEAT_WANT_CLIENT_FIRST
 	| SASL_FEAT_ALLOWS_PROXY
+	| SASL_FEAT_CHANNEL_BINDING
 	| SASL_FEAT_SUPPORTS_HTTP,	/* features */
 	gssapi_required_prompts,	/* required_prompts */
 	&gss_spnego_oid,		/* glob_context */
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 
 import argparse
+import base64
 import os
 import shutil
 import signal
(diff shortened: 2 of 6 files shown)


simo5:
However note that if I set the critical flag to 1 in the channel bindings the test fails in the client even before any communication is started.
Not clear why, help welcome.

simo5:
@quanah ping you on this one too :-)

quanah:
@simo5 Yeah, these are on my to-do list... I'm currently processing some 600+ issue reports for another project ATM.

iboukris:
As discussed, this looks generally good. However, I think it would be good to have the openldap side figured out first.
Currently the bindings passed by openldap via SASL_CHANNEL_BINDING are" tls-unique", which are good for GS2 but not for GSSAPI and would not work against Windows.
I think it would make sense to change openldap code to pass "tls-unique" for GS2 and "tls-endpoint" for GSSAPI, then it will work fine with the changes in this PR.

BTW, those are my POC branches, showing what works against Windows:
https://github.com/iboukris/openldap/commits/gssapi_cbind
https://github.com/iboukris/cyrus-sasl/commits/sasl_gssapi_cbt

quanah:
@iboukris If code changes are necessary to openldap, please file an issue on our issue tracker (bugs.openldap.org) and create a fork in our gitlab instance (git.openldap.org) to merge from, or alternatively, attach the changes in git-format-patch format to the issue in the bugzilla tracker.  Thanks!

iboukris:
@quanah I filed a bug for openldap, still working on the code :)
https://bugs.openldap.org/show_bug.cgi?id=9189

simo5:
@quanah understood, note that it would be nice to close quite a few of the issues in this repo too, some of them are so old they resolved themselves :-)

iboukris:
> I think it would make sense to change openldap code to pass "tls-unique" for GS2 and "tls-endpoint" for GSSAPI, then it will work fine with the changes in this PR.

I got the openldap side working this way, on top of this PR, however you'd notice there is a problem, when we set the bindings we don't know what is the chosen mechanism yet..

https://git.openldap.org/iboukris/openldap/-/commits/cbind_v2

simo5:
Perhaps openldap should refuse to do CBs if the list of mechs includes methods that either do not support CBs or have conflicting types of CBs ?

quanah:
Discussions about OpenLDAP and what it should or shouldn't do should be held on the openldap-devel@openldap.org list please. :)

iboukris:
@simo5 I don't see why do that, and possibly break some usage. I'm also not sure if openldap is supposed to know how to parse the mech-list, it would seem like layer violation (in fact, even the the strcmp(GSSAPI) seem unappropriated).

I think it would be simpler to just add SASL_GSSAPI_CHANNEL_BINDING, as in my previous POC branches, and let the sasl plugin decide which to use.

I'll follow up on openldap-devel@openldap.org

simo5:
@quanah the discussion is relevant here as @iboukris is questioning whether the SASL interface being used is correct

simo5:
@iboukris channel bindings require application knowledge, I would say that ultimately it is the application that needs to use the API correctly and provide the correct channel bindings, as the type of bindings is not decided by SASL but is decided by the application.
Providing a different channel binding type doesn't help as the application still need to provide the correct data in it.
The fact LDAP uses different channel binding types for different SASL mechanism *is* a problem, but it is a problem that the application drew on itself.
It shouldn't have done it in the first place.
The real issue is that tls-unique was use with good intent but was ultimately misguided because it is a broken way to deal with channel bindings (it breaks in various scenarios), which is why tls-endpoint ended up being the used one.

simo5:
@iboukris because of this layering violation anyway I think it is perfectly fine for openldap to recognize specific mechanism for which it knows it needs to use specific channel bindings.
After all mechs are a public API of SASL, so it is not like peeking into some private opaque pointer, that can change at any time.

So I think we should not introduce new options, note that the new option you proposed would be even more of a violation, because now sasl does know at all about these bindings, or we'd have to complicate the cyrus-sasl code by taking both these options in account, with some complicated matrix, what if you provide one set of CBs but not the other ?

If we do not want to detect in the openldap library what CBs to use, then I think the only other option would be to let the openldap user specify what channel bindings to use for a specific connection, and set some default. In this case as well though the interface that would be used in cyrus-sasl is the same, it would still be on openldap side to know which CB to use, only instead of embedding this knowledge in the library it would pass the buck to the user.

From SASL point of view I do not see how exploding the number of ways to pass in Channel Bindings can be a good thing at all.

iboukris:
@simo5 i hardly follow your arguments, in short, the way I suggested of adding SASL_GSSAPI_CHANNEL_BINDING, allows everything to work, does not break current usage, and does not require adding flags in openldap for it to work (which will be a burden).

hyc:
I have to agree with Simo, it looks like a code/data explosion if you have to add code to libldap for every new mechanism that wants a particular type of binding. If the hardcoded value we're currently using isn't appropriate then I guess we have to add an ldap_set_option() flag to allow it to be configured. (Of course, the value we're using now was from RFC, so if someone is violating the RFC, I'm less inclined to support their use case.)

simo5:
@hyc the real problem is that channel bindings is a SASL level concept and there is no RFC specifying what channel bindgins should be used for the LDAP protocol when using TLS that I could find.

The only RFC that mentions something is RFC 5801 that establishes the GS2 mechanism.
It says:
> A default channel binding type agreement process for all SASL application protocols that do not provide their own channel binding type agreement is provided as follows.
> 'tls-unique' is the default channel binding type for any application that doesn't specify one.

So the problem is that when channel bindings were introduced, openldap went with the generic default that GS2 specified, while Microsoft went with tls-endpoint.

There are very good operational reasons to chose tls-endpoint and Microsoft does not implement or support GS2so they wouldn't be subject to something defined for that mechanism.
At the same time there is also nothing wrong in using what an RFC defined as a default.

So the crux of the matter in the end is:
- from cyrus-sasl pov there is no conflict, because sasl is completely oblivious to the outer channel on which SASL is used, and cannot care to distinguish between mechanisms.
- from openldap pov there is a big interoperability problem that needs to be resolved, as this CB feature is critical to let openldap clients access Active Directory domains that enforce the use of Channel Bindings.

We have a few solutions, I'll present a few potential ones:

1. Do nothing
 - openldap just uses tls-unique and does not interoperate with AD servers at all

2. we decide nobody cares about GS2 (I've yet to find real world use for this mech)
 - the default used till today is changed and we make a breaking change to always use tls-endpoint regardless

3. introduce ldap_set_option() as the only way to handle this
 - clients all need to have code modifications to be able to interoperate with AD servers

4. as (3) but we change the default to tls-endpoint
 - only GS2 using applications are impacted
 - clients using GSSAPI today interoperate with AD servers by default

5. use heuristics in openldap
 - the library inspects the mechs list and if GS2 is there preserves the use of tls-unique, while if GSSAPI or GSS-SPNEGO are present the library switches to use tls-endpoint bindings.
 - it remains to be decided if tls-unique or tls-endpoint is the default for any other mechanisms, for example SCRAM uses CBs as well.

6. use heuristics *and* introduce ldap_set_option() as well
- the heuristics works as for (5), however if someone uses ldap_set_options() then heuristics are disabled and whatever is set is used.

I am currently partial to (6), as this will allow the minimal disruption to existing client code while allowing interop with Windows, it also allow applications that need to change the default for whatever reason to do it (perhaps to allow interop with less wildly popular servers that also implemented CB but using a different default for any of these sasl mechanisms).

iboukris:
Bear in mind that currently cyrus doesn't pass any bindings in GSSAPI, which is better than passing non-matching bindings (depending on the enforcement level on the server side, against Windows/Heimdal, and soon MIT as well, with basic enforcement level the bindings are only checked if the client sent them).

Changing cyrus to pass the bindings in GSSAPI and keep "tls-unique" as default binding in openldap, could cause more harm than good for AD interoperability.

> it looks like a code/data explosion if you have to add code to libldap for every new mechanism that wants a particular type of binding.

The intention wasn't to have a bindings type per mechanism, but to provide both tls-unique and tls-endpoint (RFC 5929) to cyrus, and let it decide which one to use if at all, according to the chosen mechanism.

hyc:
Is your "tls-endpoint" the same as "tls-server-endpoint" in RFC5929? It seems problematic since it only works for handshakes that used PKI server certificates. I.e., it precludes anything using anonymous Diffie-Hellman etc.

iboukris:
Yes it is.

iboukris:
> Changing cyrus to pass the bindings in GSSAPI and keep "tls-unique" as default binding in openldap, could cause more harm than good for AD interoperability.

On the server side we also need to be careful, changing cyrus to pass the bindings by default in GSSAPI will cause the server to require bindings with current MIT semantics, which will break it for clients who do not provide it (my plan was to have a set_ldap_option() for the server side).

iboukris:
> It seems problematic since it only works for handshakes that used PKI server certificates.

Right, in the above wip branch we return no bindings in that case.

On the other hand, Simo pointed out that "tls-endpoint" can be made to work with proxies, since it's just a hash of the certificate it is predictable, and the backend server can be made aware of it.

iboukris:
> On the server side we also need to be careful, changing cyrus to pass the bindings by default in GSSAPI will cause the server to require bindings with current MIT semantics, which will break it for clients who do not provide it (my plan was to have a set_ldap_option() for the server side).

@simo5 due to the above, even if openldap switches to pass "tls-endpoint" bindings overnight, I'd still be concerned to pass the bindings in GSSAPI, as it would mean the bindings are required with current MIT libs and break it for some clients.
That seems not an issue for GS2, as the client has a way to inform the server if it has sent bindings in the SASL layer, so the server can avoid passing the bindings to the GSS layer in that case.

simo5:
@iboukris I think openldap needs to make this change controlled by a configure switch/configure check, so that it starts doing it only when compile against a gssapi library that has the proper behavior, and downstream maintainer can enable it when they are ready,

iboukris:
@hyc & @simo5, how about changing openldap to not pass any bindings by default, both as client and server, and have an ldap_set_option() to send either kind of bindings (with defaults in ldap.conf).
That won't be ideal but would less likely to cause issues, and should work with this PR changes.

simo5:
@iboukris Sounds good to me.
Being controllable with ldap.conf is also a good thing.

@quanah do you think you'll have time to review and merge this PR?

hyc:
Adding a new ldap_set_option in OpenLDAP seems like the safest approach. So we will add something like LDAP_OPT_X_SASL_CBINDING, with what arguments? Pseudocode or actual patch welcome...

iboukris:
I'll work out a patch based on the above, the arguments would be "NONE, TLS_UNIQUE, TLS_ENDPOINT", with default as NONE to avoid mismatches and over enforcement.

simo5:
@iboukris @quanah are we good now to merge this one?

quanah:
Do we want to wait on the OpenLDAP side to be complete as well?

iboukris:
As long we are ok with this direction I think it is fine not to wait. I've turned it into wip MR:
https://git.openldap.org/openldap/openldap/-/merge_requests/26

simo5:
Thanks @quanah

Neustradamus:
@simo5: Good job!

Neustradamus:
@simo5, @quanah, @iboukris, @hyc: It is official for TLS 1.3 Binding!
- RFC 9266: Channel Bindings for TLS 1.3: https://tools.ietf.org/html/rfc9266

Details:
- tls-unique for TLS =< 1.2 (RFC5929)
- tls-server-end-point =< 1.2 + 1.3 (RFC5929)
- tls-exporter for TLS = 1.3 (RFC9266)

For your information, SCRAM-BIS is in progress: https://tools.ietf.org/html/draft-melnikov-scram-bis.

PR #602, opened 2020-03-25
Mechanism table update (for now, doc only proposal)

vanrein:
In a few steps, I've revised the table of authentication mechanisms.  This table was long overdue for such an update, I think.

I added columns for Post Quantum protection (which is not an issue for authentication until Quantum Computers actually arrive, unlike for encryption, but systems change slowly so this is a useful aspect to document).

I added a column for the current state according to the IANA registry of SASL mechanism names.

I could not find anything on G2, and am wondering if it might be a misspelled GS2 name?

I have removed the remark about encryption from MAX SSF, as this is not considered of value in SASL anymore; it is mostly about authentication not encryption.  I updated the description to reference brute-force search space instead, and added a value for low password quality and many-rounds effort on low password quality.  The term MAX SSF might suggest that the password quality can be 128 bit, however, which is one of many ways in which the whole MAX SSF notion is confusing and perhaps disinformation.

I edited the MAX SSF column, and am well aware that it is subjective.  Still, it did not reflect reality at all -- Kerberos5 has long deprecated DES, EXTERNAL is usually based on strong crypto, and so on.

I tried to make separately rejectable/acceptable commits out of this.  Please use that when you (dis)agree with (parts of) this proposal.  Any of these updates would improve the table, IMHO.


vanrein:
This may lead to some discussion, once the storm settles let me know and I can dive into the code to update SSF there too.

quanah:

vanrein:
I'm not sure what you are asking, or what "our DCO" means.

quanah:
> 
> 
> I'm not sure what you are asking, or what "our DCO" means.

Developer Certificate of Origin: https://developercertificate.org/

Generally, you need to add a "signed off" field to your commit (git commit -s --amend is generally sufficient) and then re-push your commit.

quanah:
Reminder that this still needs sign off in accordance with the dco to be considered.

Neustradamus:
@vanrein: Have you seen the @quanah comment?

"Generally, you need to add a "signed off" field to your commit (git commit -s --amend is generally sufficient) and then re-push your commit."

Neustradamus:
@vanrein: Have you progressed on your PR?
Already 5 years, 5 months.
Have you seen the @quanah comments?

Issue #378, opened 2013-03-21
Java JNDI Client Connects to OpenLDAP fail


brong:
From: [Bill MacAllister](whm@stanford.edu)
Bugzilla-Id: 3787
Version: 2.1.25
Owner: [Alexey Melnikov](Alexey.Melnikov@isode.com)

quanah:
This might be the same as issue #419 

As I recall when I worked @ Stanford, JNDI defaults to having an SSF of 0.  It was one of the reasons we dumped JNDI for UnboundID.  I'm surprised anyone @ Stanford was continuing to use JNDI given its many flaws.

quanah:
@whm this is you IIRC. ;)

whitehse:
@whm

Issue #419 has been resolved in 2.1.27 rc6. Can you retest this, or provide instructions for reproducing your test?

quanah:
@whm no longer works at Stanford, and may not have any ability to test this.  I still have connections at Stanford, so I can see if any of them can test this.

quanah:
Going to close this, I believe this is fixed with #419 being resolved.

If someone continues to have this issue, please reopen.

PR #603, opened 2020-04-15
Allow GSS-SPNEGO to negotiate a ssf of 0

simo5:
In GSS-SPNEGO ssf negotiation is implicit and depends on GSSAPI level Integrity and Confidentiality flags.
In order to interoperate with Windows, which does not allow nested security layers, we need to be able to set ssf to 0 when GSS-SPNEGO is negotiated within an outer TLS channel.
Unfortunately MIT's GSSAPI always forces negotiation of Integrity and Confidentiality regardles of what flags are passed to gss_init_sec_context()
Add code to allow MIT to skip setting those flags on the wire as Windows interprets those as a *request* rather than an offer like MIT does.

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -2123,7 +2123,50 @@ static int gssapi_client_mech_step(void *conn_context,
 		/* We want to try for privacy */
 		req_flags |= GSS_C_CONF_FLAG;
 	    }
-	}
+#ifdef HAVE_GSS_KRB5_CRED_NO_CI_FLAGS_X
+        /* The krb5 mechanism automatically adds INTEG and CONF flags even when
+         * not specified, this has the effect of rendering explicit requests
+         * of no confidentiality and integrity via setting maxssf 0 moot.
+         * However to interoperate with Windows machines it needs to be
+         * possible to unset these flags as Windows machines refuse to allow
+         * two layers (say TLS and GSSAPI) to both provide these services.
+         * So if we do not suppress these flags a SASL/GSS-SPNEGO negotiation
+         * over, say, LDAPS will fail against Windows Servers */
+	} else if (params->props.max_ssf == 0) {
+            gss_buffer_desc empty_buffer = GSS_C_EMPTY_BUFFER;
+            if (client_creds == GSS_C_NO_CREDENTIAL) {
+                gss_OID_set_desc mechs = { 0 };
+                gss_OID_set desired_mechs = GSS_C_NO_OID_SET;
+                if (text->mech_type != GSS_C_NO_OID) {
+                    mechs.count = 1;
+                    mechs.elements = text->mech_type;
+                    desired_mechs = &mechs;
+                }
+
+                maj_stat = gss_acquire_cred(&min_stat, GSS_C_NO_NAME,
+                                            GSS_C_INDEFINITE, desired_mechs,
+                                            GSS_C_INITIATE,
+                                            &text->client_creds, NULL, NULL);
+                if (GSS_ERROR(maj_stat)) {
+                    sasl_gss_seterror(text->utils, maj_stat, min_stat);
+                    sasl_gss_free_context_contents(text);
+                    return SASL_FAIL;
+                }
+                client_creds = text->client_creds;
+            }
+
+            maj_stat = gss_set_cred_option(&min_stat, &client_creds,
+                                           (gss_OID)GSS_KRB5_CRED_NO_CI_FLAGS_X,
+                                            &empty_buffer);
+            if (GSS_ERROR(maj_stat)) {
+                sasl_gss_seterror(text->utils, maj_stat, min_stat);
+                sasl_gss_free_context_contents(text);
+                return SASL_FAIL;
+            }
+#endif
+        }
+
+
 #ifdef GSS_USE_CCACHE_STORE
         /* TODO: This should maybe depend on "pass_credentials" (server-side), or Kerberos token attributes */
         params->props.security_flags |= SASL_SEC_PASS_CREDENTIALS;
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -2113,9 +2113,15 @@ static int gssapi_client_mech_step(void *conn_context,
 
 	/* Setup req_flags properly */
 	req_flags = GSS_C_MUTUAL_FLAG | GSS_C_SEQUENCE_FLAG;
-	if (params->props.max_ssf > params->external_ssf) {
-	    /* We are requesting a security layer */
+        if (text->mech_type == NULL) {
+            /* in SASL/GSSAPI RFC4752 requires the INTEG flag MUST be set to
+             * TRUE in all cases */
 	    req_flags |= GSS_C_INTEG_FLAG;
+        }
+	/* We are requesting a security layer */
+	if (params->props.max_ssf > params->external_ssf) {
+            /* Add integrity if explicity requested */
+            req_flags |= GSS_C_INTEG_FLAG;
 	    /* Any SSF bigger than 1 is confidentiality. */
 	    /* Let's check if the client of the API requires confidentiality,
 	       and it wasn't already provided by an external layer */
@@ -2132,7 +2138,7 @@ static int gssapi_client_mech_step(void *conn_context,
          * two layers (say TLS and GSSAPI) to both provide these services.
          * So if we do not suppress these flags a SASL/GSS-SPNEGO negotiation
          * over, say, LDAPS will fail against Windows Servers */
-	} else if (params->props.max_ssf == 0) {
+	} else if (params->props.max_ssf <= 1) {
             gss_buffer_desc empty_buffer = GSS_C_EMPTY_BUFFER;
             if (client_creds == GSS_C_NO_CREDENTIAL) {
                 gss_OID_set_desc mechs = { 0 };
--- a/m4/sasl2.m4
+++ b/m4/sasl2.m4
@@ -289,6 +289,19 @@ if test "$gssapi" != no; then
 
   cmu_save_LIBS="$LIBS"
   LIBS="$LIBS $GSSAPIBASE_LIBS"
+  if test "$ac_cv_header_gssapi_gssapi_krb5_h" = "yes"; then
+    AC_CHECK_DECL(GSS_KRB5_CRED_NO_CI_FLAGS_X,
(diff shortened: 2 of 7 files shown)


simo5:
@quanah PTAL this is a follow up on the Channel Bindings stuff merged last week

simo5:
@iboukris PTAL

iboukris:
I wonder if the change should be limited to GSS-SPNEGO mech. Other than that it looks fine and solves the problem at hand.

simo5:
No, the change should be available to both, it is critical for GSS-SPNEGO interoperability, but there is no point in restricting it to GSS-SPNEGO only.

iboukris:
Yeah, that's fine too.

quanah:
just a note, I'm generally holding off on new feature work until the project has an established contribution/licensing set of guidelines in place.. I will continue to look through minor bug fix requests though.

iboukris:
@simo5 actually per my reading of rfc4752, this should really be done only for GSS-SPNEGO mech, while for GSSAPI mech we should always ask for 'integ_req_flag' as i think it is needed to protect the SSF negotiation (I guess that's why Windows allows such ap-req even over TLS for the GSSAPI mech).
Quote: When calling the GSS_Init_sec_context, the client MUST pass the integ_req_flag of TRUE

iboukris:
Although technically it looks like it works even when no integ/conf flags are requested in the ap-req even for the GSSAPI mech, and the SSF neg seem to work fine in that case with MIT libs (that is the calls to gss_wrap/unwrap).

iboukris:
Actually, if we would always explicitly request GSS_C_INTEG_FLAG for the GSSAPI mech, then the GSS_KRB5_CRED_NO_CI_FLAGS_X doesn't seem to matter and the flags are requested anyway, but I still think we shouldn't request GSS_KRB5_CRED_NO_CI_FLAGS_X for GSSAPI mech.

simo5:
Unlike in Windows' SSPI, in GSSAPI these flags are advisory, so I think the current code is probably ok.

iboukris:
Windows also seem to have no problem to wrap/unwrap in that case, see my comment on krbdev from this morning:
http://mailman.mit.edu/pipermail/krbdev/2020-June/013318.html

simo5:
Then I think we are good.

iboukris:
I think we should change the code to always request GSS_C_INTEG_FLAG for GSSAPI mech as per RFC, and then the GSS_KRB5_CRED_NO_CI_FLAGS_X wouldn't really matter. But none of it seems critical otherwise.

simo5:
@iboukris, I think this will complicate things ...

quanah:
> 
> 
> I think we should change the code to always request GSS_C_INTEG_FLAG for GSSAPI mech as per RFC, and then the GSS_KRB5_CRED_NO_CI_FLAGS_X wouldn't really matter. But none of it seems critical otherwise.

Let's be sure and avoid re-creating #419, #378 when poking at this.

simo5:
@quanah please check the added patch,
I have no tested it against wIndows though.
I am willing to drop it if *reality* rather than RFC compliance says we should

plugins/gssapi.c
```diff
@@ -1772,10 +1772,18 @@ static int gssapi_client_mech_step(void *conn_context,
 	}
 
 	/* Setup req_flags properly */
-	req_flags = GSS_C_MUTUAL_FLAG | GSS_C_SEQUENCE_FLAG;
+        if (text->mech_type == NULL) {
+            /* in SASL/GSSAPI RFC4752 requires rthe INTEG flag MUST be set to
```
iboukris, line 1776: typo

plugins/gssapi.c
```diff
@@ -1792,7 +1800,7 @@ static int gssapi_client_mech_step(void *conn_context,
          * two layers (say TLS and GSSAPI) to both provide these services.
          * So if we do not suppress these flags a SASL/GSS-SPNEGO negotiation
          * over, say, LDAPS will fail against Windows Servers */
-	} else if (params->props.max_ssf == 0) {
+	} else if (params->props.max_ssf <= 1) {
```
iboukris, line 1803: I don't see why is this change needed.

plugins/gssapi.c
```diff
@@ -1772,10 +1772,18 @@ static int gssapi_client_mech_step(void *conn_context,
...
+        if (text->mech_type == NULL) {
+            /* in SASL/GSSAPI RFC4752 requires rthe INTEG flag MUST be set to
+             * TRUE in all cases */
+	    req_flags = GSS_C_INTEG_FLAG;
+        }
 	if (params->props.max_ssf > params->external_ssf) {
 	    /* We are requesting a security layer */
-	    req_flags |= GSS_C_INTEG_FLAG;
+	    req_flags |= GSS_C_MUTUAL_FLAG | GSS_C_SEQUENCE_FLAG;
+            if (text->mech_type == &gss_spnego_oid) {
```
iboukris, line 1783: I would skip the condition and just add the flag.

iboukris:
@simo5 this currently doesn't work, if we omit the 'mutual' flag then we also need to expect no ap-rep from the server.

```
$ ldapsearch -h adc.acme.com -b dc=acme,dc=com cn=administrator -Y GSSAPI -N -O maxssf=0 -ZZ
Password for Administrator@ACME.COM: 
SASL/GSSAPI authentication started
ldap_sasl_interactive_bind: Local error (-2)
	additional info: SASL(-1): generic failure: GSSAPI Error: A required input parameter could not be read (Unknown error)

**Comment by simo5** [comment]
@iboukris ok, let me back away that part of the change

**Comment by iboukris** [comment]
@simo5 this patch solves it without the mutual flag (not pretty though): https://github.com/iboukris/cyrus-sasl/commit/611cda20874320153b9739592de5e26316c4be23

File: plugins/gssapi.c (hunk 2 above)
Comment (simo5) [reply plugins/gssapi.c:1803]: Because krb5 sets both confidentiality and Integrity flags, we still need to remove the confidentiality flag from the defaults when we are asking just integrity (ssf = 1) not only when we require no security layer at all

File: plugins/gssapi.c (hunk 1 above)
Comment (simo5) [reply plugins/gssapi.c:1777]: ok

File: plugins/gssapi.c (hunk 3 above)
Comment (simo5) [reply plugins/gssapi.c:1783]: ok

**Comment by iboukris** [comment]
Now that I read #419, I think we can really leave the 'mutual' flag alone:
https://lists.andrew.cmu.edu/pipermail/cyrus-sasl/2015-April/002809.html

@quanah I think it is possible the interop issue with the previous change was due to the mutual flag, so if we leave it alone we should be fine (at least this error looks like mine https://github.com/cyrusimap/cyrus-sasl/issues/419#issuecomment-365344413).

Regarding GSS_C_INTEG_FLAG, i think the rational to always request it, is in order to be able to call gss_wrap/unwrap to protect the SSF neg, but as I mentioned in practice it doesn't seem to be required and gss_wrap/unwrap still work.

File: plugins/gssapi.c (hunk 2 above)
Comment (iboukris) [reply plugins/gssapi.c:1801]: Oh I see, although if it doesn't solve any real problem I wouldn't bother.

Looking again at the code however, I think you can change it to simple "} else {" without any condition.

**Comment by iboukris** [comment]
I should point out that the interop issue with the 'mutual' flag change, may be a Windows specific thing. The RFC doesn't sound like an empty token is expected by the client in this case.

File: plugins/gssapi.c (hunk 2 above)
Comment (iboukris) [reply plugins/gssapi.c:1801]: Another simple way could be to always acquire creds and to always set GSS_KRB5_CRED_NO_CI_FLAGS_X, since that's what we want.  I know we had issues with acquiring creds but as long as we pass GSS_C_NO_NAME it should be equivalent, and we'd need it for other features.

**Comment by pzirnik** [comment]
Any progress here ?

**Comment by quanah** [comment]

**Comment by simo5** [comment]
Hi @quanah the commits already have a sign-off-by line, do you require anything else?
Do you have a pointer to the DCO ?

**Comment by jdehaan** [comment]
@simo5 there is a CI check for that after I rebased my PRs I could see what the CI meant was wrong with the PR and there was a hint inside the github action check on how to perform the "fix"

**Comment by quanah** [comment]
> 
> 
> Hi @quanah the commits already have a sign-off-by line, do you require anything else?
> Do you have a pointer to the DCO ?

See https://github.com/probot/dco

**Comment by simo5** [comment]
Ok, I'll rebase on master and bump to let the bot have their day.
Thanks.

**Comment by quanah** [comment]
@simo5 Please rebase to work against current master, thanks!

**Comment by simo5** [comment]
done

File: plugins/gssapi.c [hunk 4]
```diff
@@ -2113,17 +2113,66 @@ static int gssapi_client_mech_step(void *conn_context,
...
-	}
+#ifdef HAVE_GSS_KRB5_CRED_NO_CI_FLAGS_X
+        /* The krb5 mechanism automatically adds INTEG and CONF flags even when
+         * not specified, this has the effect of rendering explicit requests
+         * of no confidentiality and integrity via setting maxssf 0 moot.
+         * However to interoperate with Windows machines it needs to be
+         * possible to unset these flags as Windows machines refuse to allow
+         * two layers (say TLS and GSSAPI) to both provide these services.
+         * So if we do not suppress these flags a SASL/GSS-SPNEGO negotiation
+         * over, say, LDAPS will fail against Windows Servers */
```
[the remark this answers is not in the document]
simo5: This is a well know interop issue with AD.

plugins/gssapi.c
```diff
@@ -2113,17 +2113,66 @@ static int gssapi_client_mech_step(void *conn_context,
 
 	/* Setup req_flags properly */
 	req_flags = GSS_C_MUTUAL_FLAG | GSS_C_SEQUENCE_FLAG;
-	if (params->props.max_ssf > params->external_ssf) {
-	    /* We are requesting a security layer */
+        if (text->mech_type == NULL) {
+            /* in SASL/GSSAPI RFC4752 requires the INTEG flag MUST be set to
+             * TRUE in all cases */
```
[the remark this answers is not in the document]
simo5: Because it is not set when maxssf = 0 or no security is requested at all.

That "all cases" means IIRC all cases when any protection is requested maxssf >= 1

Ie you are not allowed to request only Confidentiality, you must always request integrity when Confidentiality is requested.

plugins/gssapi.c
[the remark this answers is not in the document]
hyc: OK just for future reference, setting maxssf=0 works with GSSAPI mech already, but without this patch GSS-SPNEGO does not.

PR #611, opened 2020-05-07
Increase MAX_REQ_LEN to allow transport of more sophisticated credent…

mdoggydog:
…ials

The constant MAX_REQ_LEN, which controls the maximum size of a password
accepted by saslauthd, is increased from 256 to 4096.  Though a simple
human-typed plaintext password is unlikely to ever be so big, this
allows saslauthd to accept other, larger credentials.  In particular,
this is large enough to allow saslauthd to receive SAML2 security
assertions and pass them through to a PAM authenticator for verification.

MAX_REQ_LEN also affects the fixed-size buffers allocated for "login",
"service", "realm", and "client_addr", though only "password" needs
to be enlarged for this use-case.  However, at this point it doesn't
seem to be worth the effort/complexity to introduce a separate constant
to independently control the size of the "password" buffer.

### DIFF
--- a/saslauthd/saslauthd-main.h
+++ b/saslauthd/saslauthd-main.h
@@ -82,7 +82,7 @@
 #define DOOR_FILE		"/mux"              
 
 /* login, pw, service, realm buffer size */
-#define MAX_REQ_LEN		256     
+#define MAX_REQ_LEN		65536
 
 /* socket backlog when supported */
 #define SOCKET_BACKLOG  	32


flowerysong:
> The constant MAX_REQ_LEN, which controls the maximum size of a password
> accepted by saslauthd, is increased from 256 to 4096. Though a simple
> human-typed plaintext password is unlikely to ever be so big, this
> allows saslauthd to accept other, larger credentials. In particular,
> this is large enough to allow saslauthd to receive SAML2 security
> assertions and pass them through to a PAM authenticator for verification.

This feels like a use case that could very easily break with this arbitrary limit; I happen to have a random SAML assertion lying around from some debugging I was doing recently and it's far more than 4096 characters:

```
ironbull-zeke ~ $ wc -c saml
20289 saml
ironbull-zeke ~ $ base64 --decode saml  | wc -c
15215
```

mdoggydog:
Yeah... 4096 was pretty much derived from "2-3x as a big as anything I had encountered in my particular use case".  I admit I don't have a good sense of how large things like SAML assertions can get out in the wild.

quanah:

mdoggydog:
> Please update your commit(s) to be signed off on in accordance with our DCO. Thanks!

Done!

(I also went and upped the size of the MAX_REQ_LEN increase, in response to the first comment.)

quanah:
tyvm!

quanah:
saslauthd is only meant for use with simple passwords, so increasing the size here is incorrect.  Ticket based mechanisms should not be using saslauthd at all.

PR #613, opened 2020-05-24
Add LOGIN to the set of static plugins built in Win32 “core” configuration

bengotow:
Hey folks,

There are several Visual Studio projects that can be used to build sasl2 on Windows, and the "core" project statically links the common mechanisms necessary for a basic client application (PLAIN, DIGESTMD5, etc). However, this default set does not include `LOGIN`, which I found out the hard way when I rebuilt sasl2 and libetpan against a newer version of OpenSSL and shipped them in an update to Mailspring last week (Related mess: https://github.com/Foundry376/Mailspring/issues/1996).

After reviewing the Visual Studio projects, I think that the exclusion of LOGIN may have been a mistake - adding it works fine and the code is structurally similar to the PLAIN mechanism. I've made the necessary changes to the Visual Studio projects below and tested the resulting sasl2.dll.


quanah:

hyc (approved):

Issue #319, opened 2009-11-15
realm specific configuration for LDAP


brong:
From: [John Newbigin](jnewbigin@swin.edu.au)
Bugzilla-Id: 3181
Version: 2.1.x
Owner: [Ken Murchison](murch@andrew.cmu.edu)

jdehaan:
I think this is a really great feature that avoids many quirky workarounds in order to get the same benefits like described here: https://ltb-project.org/documentation/general/sasl_delegation

I do not know if the potential connection caches are ok with that though... These would have to be maintained per realm as well (I am unfamiliar with the internals of the daemon, I am unsure if this is really missing or if there is no cache then also no problem)

hyc:
See back-meta solution in comments of #621

PR #621, opened 2020-08-07
Feature #319: support multiple ldap realms

jdehaan:
Feature #319, cleanup of the original patch from John Newbigin

**Changes**:

- be able to handle several LDAP realm configurations within one daemon setup
- adds the keyword `realm` that can be used to delimit several sets of the standard parameters within the conf file into sections (one section associated with each realm) 

This avoids ugly workarounds in practical setups and allows to use saslauthd as a "proxy" for several **different** LDAP realms. NOTE that this differs from the multiple servers setup in the `ldap_servers` for redundancy/failover.

Another point is that there are different use cases you can get into the code. One is via `sasl_checkpasswd` API call without a realm and via the mutex and providing the realm (what I tested initially only and figured out openldap uses the API call instead). So during the code review some eye on how the realm is treated is needed as well.

CACHE:

* LAK cache via static variable removed, because the configuration is now dynamic / dependent on parameters. An alternative implementation could load all the settings in memory and store the realms in a table (not done).
* Is the auth cache still ok with the changes? The cache handling is also to be checked the realm shall always be part of the key for the cache. As far as I tested looks like it behaves correctly.

**KNOWN ISSUE**:

There seems to be an issue with the certificate verification in multiple realm mode. The child process seems to remember the TLS CA CERT file to use and uses randomly (depending on which worker process was selected) a wrong one. Not sure where this is persisted and how to remove it yet. Can be worked around by providing a global `ldap_tls_cacert_file` setting with all CA in it.

### DIFF
--- a/saslauthd/auth_ldap.c
+++ b/saslauthd/auth_ldap.c
@@ -61,19 +61,20 @@ auth_ldap(
   /* END PARAMETERS */
   )
 {
-	static LAK *lak = NULL;
+	LAK *lak = NULL;
 	int rc = 0;
 
-	if (lak == NULL) {
-		rc = lak_init(SASLAUTHD_CONF_FILE, &lak);
-		if (rc != LAK_OK) {
-			lak = NULL;
-			RETURN("NO");
-		}
+	rc = lak_init(SASLAUTHD_CONF_FILE, &lak);
+	if (rc != LAK_OK) {
+		lak = NULL;
+		RETURN("NO");
 	}
 
 	rc = lak_authenticate(lak, login, service, realm, password);
-    	if (rc == LAK_OK) {
+
+	lak_close(lak);
+
+	if (rc == LAK_OK) {
 		RETURN("OK");
 	} else {
 		RETURN("NO");
--- a/saslauthd/lak.c
+++ b/saslauthd/lak.c
@@ -1577,15 +1577,26 @@ int lak_authenticate(
 	if (EMPTY(user))
 		return LAK_FAIL;
 
+	/* use the specified realm from the user (@realm suffix) */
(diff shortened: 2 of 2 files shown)


quanah:

hyc:
> KNOWN ISSUE:

>There seems to be an issue with the certificate verification in multiple realm mode. The child process seems to remember the TLS CA CERT file to use and uses randomly (depending on which worker process was selected) a wrong one. Not sure where this is persisted and how to remove it yet. Can be worked around by providing a global ldap_tls_cacert_file setting with all CA in it.

If that's the case, you probably need to move all of the TLS settings in lak_connect() to after the ldap_initialize(), and just set them on the lak->ld. Since you're dealing with different settings for different realms, should avoid doing any global LDAP settings at all, and all of them should just be on the lak->ld handle instead.

hyc (changes requested):
See comment https://github.com/cyrusimap/cyrus-sasl/pull/621#issuecomment-1610279518

jdehaan:
You're fully right 🤦🏻 ... This was not needed in out case because of the special corner case that we had the same root certificate... In the end all this looks like supporting multiple reals is a use case not at all foreseen by the initial design of the tool. It might be wiser to close this PR unmerged after all since there are many difficulties to overcome left over and there is (I discovered that a bit late) a cleaner possible - but not straight-forward - setup using a meta backend:

https://ltb-project.org/documentation/sasl_delegation.html#pass-through-authentication-on-several-ldap-directories-with-openldap-meta-backend

image

Thanks for your time and thorough review!

hyc:
Yeah, closing this in favor of using back-meta makes sense.

PR #622, opened 2020-09-05
respect namelen in get_fqhostname

ilejn:
The reason of the PR is memory corruption in get_fqhostname.
name[namelen] = '\0' is effectively out of bounds if namelen is sizeof, which is the case.
The issue introduced in cafd436fdb7bb79554b9564acf8345d21e25604d


quanah:

hyc (approved):

PR #638, opened 2021-01-13
Remove duplicated md5global header

theidexisted:
These three files are identical except for some space indention in the comment, so I think we can simplify it.


quanah:
This merge request needs updating as some of the files in question no longer exist.

Please update your commit(s) to been signed off on in accordance with our DCO as well.  Thanks!

quanah:
closing in preference of #664

Issue #624, opened 2020-09-11
Build failure with sphinx 3.2.1


panlinux:
cyrus-sasl2 ships with a sphinx extension to build its documentation, and this extension was based on a very old sphinx version. It no longer builds with sphinx 3.2.1, failing in several places. Here are the first two:

a) "NoUri"
Extension error:
Could not import extension sphinxlocal.builders.manpage (exception: cannot import name 'NoUri' from 'sphinx.environment' (/usr/lib/python3/dist-packages/sphinx/environment/__init__.py))
make[2]: *** [Makefile:1166: man/.sphinx-build.stamp] Error 2
make[2]: Leaving directory '/home/ubuntu/deb/cyrus-sasl2-2.1.27+dfsg/build-heimdal'
make[1]: *** [Makefile:686: all-recursive] Error 1
make[1]: Leaving directory '/home/ubuntu/deb/cyrus-sasl2-2.1.27+dfsg/build-heimdal'
make: *** [Makefile:556: all] Error 2

A grep shows `NoUri` is imported but apparently never used, so it could be dropped.

b) MACRO_DEF
Extension error:
Could not import extension sphinxlocal.builders.manpage (exception: cannot import name 'MACRO_DEF' from 'sphinx.writers.manpage' (/usr/lib/python3/dist-packages/sphinx/writers/manpage.py))
make[4]: *** [Makefile:1166: man/.sphinx-build.stamp] Error 2
make[4]: Leaving directory '/home/ubuntu/git/packages/cyrus-sasl2/cyrus-sasl2/build-heimdal'
make[3]: *** [Makefile:686: all-recursive] Error 1
make[3]: Leaving directory '/home/ubuntu/git/packages/cyrus-sasl2/cyrus-sasl2/build-heimdal'
make[2]: *** [Makefile:556: all] Error 2
make[2]: Leaving directory '/home/ubuntu/git/packages/cyrus-sasl2/cyrus-sasl2/build-heimdal'
dh_auto_build: error: cd build-heimdal && make -j4 sasldir=/usr/lib/x86_64-linux-gnu/sasl2 returned exit code 2
make[1]: *** [debian/rules:164: override_dh_auto_build] Error 25
make[1]: Leaving directory '/home/ubuntu/git/packages/cyrus-sasl2/cyrus-sasl2'
make: *** [debian/rules:122: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2

[...]

PR #640, opened 2021-02-07
Fix Sphinx errors

jspricke:


quanah:
This seems like you're asking us to commit someone else's work in addition to your own?  Both authors would need to sign off on the commit for our DCO, thanks.

jspricke:
@panlinux I used your patch from #624, could you send a `Signed-off-by` version of it?

panlinux:
I created a new branch, converted that diff from the comment into a commit, and added the signed-of. Is this ok now?

Branch: https://github.com/panlinux/cyrus-sasl/tree/partial-fix-issue-624
Commit: https://github.com/panlinux/cyrus-sasl/commit/9f90ba7e26c38118cbd26e21a003327ef6769578

quanah:
Yes, but you'll have to create a new pull request for your new branch and close this one I guess :)

quanah:
or we need that commit pulled into this pull request or something ;)

jspricke:
Thanks Andreas, All commits are signed off now.

PR #642, opened 2021-03-03
configure: fix check for dlsym underscore

nacho:
The exit function requires to include stdlib otherwise
this will fail on new versions of MacOS


quanah:
Please update your pull request so it has been signed off on in accordance with our DCO.  Thanks!

PR #648, opened 2021-05-09
saslauthd: Fix saslauthd.conf path in manpage.

Biswa96:


quanah:
Rather than using a hard coded path, this issue should be addressed by doing a install time substitution based on the SASLAUTHD_CONF_FILE_DEFAULT variable.

PR #653, opened 2021-06-21
OpenSSL 3 fixes

simo5:
In OpenSSL 3.0 these algorithms have been moved to the legacy provider
which is not enabled by default. This means allocation can and do fail.
Handle failed allocations by returning an actual error instead of
crashing later with a NULL context.

Also deal with deprecation warnings about HMAC()

### DIFF
--- a/plugins/scram.c
+++ b/plugins/scram.c
@@ -65,7 +65,9 @@
 
 #include <openssl/sha.h>
 #include <openssl/evp.h>
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
 #include <openssl/hmac.h>
+#endif
 
 /*****************************  Common Section  *****************************/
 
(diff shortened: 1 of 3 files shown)


quanah:
Shouldn't we be looking at supporting OpenSSL 3.0.0? Perhaps an issue on that would be good.

simo5:
We compiled cyrus-sasl with openssl 3.0, and this is the first thing we see with openldap clients, DIGEST-MD5 failing.
I am ok if you want to cretae  amilestone for formal openssl 3.0 support and add this PR there.
We'll send more patches as our testing with OpenSSL 3.0 continues if there are other issues.

simo5:
@quanah, I just realized that we do not seem to run make check in github actions or similar, but that would definitely be good to make sure patches like this actually pass build before merge. Is that something we should enable?

quanah:
> 
> 
> @quanah, I just realized that we do not seem to run make check in github actions or similar, but that would definitely be good to make sure patches like this actually pass build before merge. Is that something we should enable?

I asked for CI/CD to be renabled, since I don't have that level of control and haven't heard back.  I'll ping again.

simo5:
Adding more OpenSSL fixes

simo5:
Added a patch to deal with deprecation warnings about HMAC(), there are more warnings for DES() functions as well, we'll see if I have the b/w to deal with that later

plugins/scram.c
```diff
@@ -289,6 +293,63 @@ print_hash (const char * func, const char * hash, size_t hash_size)
 }
 #endif
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+
+/* Decalre as void given functions never use the result */
+void *HMAC(const EVP_MD *evp_md, const void *key, int key_len,
+                     const unsigned char *data, size_t data_len,
+                     unsigned char *md, unsigned int *md_len)
+{
```
beldmit, line 302: I'd use EVP_Q_mac function for this purpose

Neustradamus:
@quanah, @hyc: What do you think?
A merging before the 2.1.28?

simo5:
Fixed a memleak found by covscan and rebased on latest master.

hyc:
Everything looks good here. As a side question - it's all using the default libctx. Any thoughts on how to let users specify non-default providers?

simo5:
@hyc are you asking in general or are asking in the context of #668 which introduces custom code for loading the legacy provider?

hyc:
In general, since I presume we should have the ability to use custom PKCS#11 modules, etc.

simo5:
In general an application can load additional providers on the default OpenSSL context (NULL) and even change OpenSSL configuration to prefer a provider over another.
I wanted to do the legacy loading in the other PR by copying the default context and then just adding the legacy provider exactly to maintain that semantic, unfortunately that is not really possible with 3.0 APIs. See https://github.com/openssl/openssl/issues/16728

hyc (approved):

PR #654, opened 2021-06-26
SCRAM Fix: Exception when server does not support channel binding

GuidoKiener:
When server does not support channel binding, but client requires
channel binding with the second message, then the server crashes with
an exception in strcmp(sparams->cbinding->name, ...)


Neustradamus:
@GuidoKiener: Nice to see new improvements of SCRAM into Cyrus SASL.

Neustradamus:
Thanks @GuidoKiener, @hyc, @quanah!

PR #660, opened 2021-09-01
Fix iovec name conflict

kovdan01:
Cyrus SASL is not the only library that defined struct iovec when it is not
provided by OS interface (for example, on Windows) -  msgpack-c does the same:
github.com/msgpack/msgpack-c/blob/6e7deb8/include/msgpack/vrefbuffer.h#L16
To avoid name conflicts because of redefinition of struct iovec when using both
libraries in the same project, this commit proposes cyrus_sasl_iovec typedef.

### DIFF
--- a/common/plugin_common.c
+++ b/common/plugin_common.c
@@ -171,7 +171,7 @@ int _plug_ipfromstring(const sasl_utils_t *utils, const char *addr,
     return SASL_OK;
 }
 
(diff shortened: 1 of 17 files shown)


kovdan01:
Not relevant anymore - msgpack-c already fixed it in latest release: https://github.com/msgpack/msgpack-c/releases/tag/c-4.0.0

PR #666, opened 2021-10-02
Fix windows makefiles

SyntevoAlex:
This series of patches fix all build errors I have encountered on Windows.

Please refer to commit messages.
Please refer to last commit for instructions on how to build on Windows.


SyntevoAlex:
@Ri0n you could be interested

Ri0n:
LGTM.
I haven't touched the project quite a while but to be honest I'd better put my efforts into CMake build scripts implementation.

So what about static plugins?

SyntevoAlex:
CMake would be nice, but that's going to be a latger project and unfortunately I can't spend so much time.
Static plugins are not supported - they were never supported in `NTMakeFile` build and supporting them is also some effort.

Ri0n (approved):

SyntevoAlex:
Thanks for having a look!

SyntevoAlex:
I don't think that any of my changes affect cygwin build. Instead, they are targeted at `NTMakeFile` build.
I have removed cygwin from docs because it's no longer required, the new PowerShell step can be used in both cases.

quanah:
Then the documentation should note either can be used to build cyrus-sasl on Windows.

SyntevoAlex:
That's not true because the documented cygwin step is also broken :)

quanah:
@SyntevoAlex Is there a missing file in these commits, for plugins/makeinit.ps1 ?

quanah:
commits go into master first, then into 2.1 (I just merged it there)

Issue #665, opened 2021-09-30
Des/Rc4 implementation in digestmd5 and deprecations.


simo5:
digestmd5 defined 3 possible choices for rc4: rc4-40, rc4-56 and rc4, however as far as I can tell they all end up using the same functions which are not parametric (the functions only implement the full strength RC4).

Additionally there is a full custom re-implementation of RC4 in the file in addition to the code calling openssl.

On the DES side digestmd5 still defines des and 3des as options.
Here there are also 2 implementations, custom and openssl basedof des, on which 3des is constructed.

All this is old cruft that is cryptographically quite broken.

Proposal:
- remove the custom implementations and always depend on openssl's legacy provider.
- drop the rc4-40,rc4-56 defintions as they are useless.
- drop support for classic des and declare 3des deprecated.

Finally declare the whole module deprecated, and prepare for dropping it in some future release.

quanah:
Probably declare as deprecated in 2.1.28, remove in 2.2

simo5:
is it ok to drop the custom implementations in 2.1.28 and always rely on openssl ?

quanah:
@hyc thoughts on @simo5 question? :)

hyc:
Alll of that sounds good to me. There's no good reason to keep a homegrown crypto implementation, especially when we already depend on OpenSSL.

hyc:
As usual for deprecating/removing something like digestmd5, the replacement (SCRAM) should be in wide use before the actual deletion/removal.

simo5:
agreed!
I have some patches almost ready, will open a PR, it is a follow up of my latest one to enable building on openssl3

Neustradamus:
@quanah, @hyc: What do you think?
It will be good before the 2.1.28 release?

bgermann:
This should be removed from the 2.2.0 milestone in favor of #751.

bgermann:
As far as I can see, the whole module was deprecated with b85268ec0f1ace84b83b45cf8385f978da0c19cc which was released with 2.1.27. So why should it not be removed in 2.2 as RFC6331 also declared it historic?

bgermann:
The removal is implemented with  #773.

quanah:
[...]

PR #668, opened 2021-10-05
Load OpenSSL 3 legacy provider explicitly in digestmd5 plugin

simo5:
Add support to load legacy provider directly in digestmd5 plugin so that rc4 is available.
This depends on #653 to be pushed first as it includes the same commits.
Addresses some of the points in #665


simo5:
Marking as WIP because I do not have a test that exercises this plugin, to assure it works as desired

simo5:
This seems to be working fine in Fedora Rawhide

Neustradamus:
@quanah, @hyc: What do you think?
A merging before the 2.1.28?

simo5:
Incorporated changes on #653 and therefore also rebased on master

hyc (changes requested):
relying on OpenSSL's RC4 seems to be going the wrong direction, since it's optional there and frequently not built. Doing this properly requires adding configure tests for OpenSSL's RC4 support. That seems like a bunch of unnecessary hassle. Unless you know of actual flaws in the RC4 impl here, I'd drop this commit.

simo5:
The reason to do this is that we actually want RC4 to be disable-able if needed in the distribution, and that is easier to do in one place (openssl) than trying to chase many different components.
I can change the code if you think we can pursue, but if you really want to drop it let me know, and I'll just maintain a downstream patch for it for now.

quanah:
Need to see if there are any mechanisms for 2.2 that will use RC4.  If not, this PR can be closed.

plugins/digestmd5.c
```diff
@@ -1102,60 +1114,139 @@ static int init_des(context_t *text,
...
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+    ossl3_context_t *ossl3_ctx;
 
-    rc = EVP_EncryptInit_ex(ctx, EVP_rc4(), NULL, enckey, NULL);
-    if (rc != 1) return SASL_FAIL;
+    rc = init_ossl3_ctx(text);
+    if (rc != SASL_OK) return rc;
+
+    ossl3_ctx = (ossl3_context_t *)text->crypto.libctx;
+    cipher = EVP_CIPHER_fetch(ossl3_ctx->libctx, "RC4", "");
```
mistotebe, line 1211: `cipher` is leaked if we got it from here as seen in openldap's test076-authid-rewrite (just build with `-fsanitize=address` or run the first `ldapwhoami` under valgrind).

simo5: nice catch

simo5:
@quanah should we just close this PR ?

plugins/digestmd5.c
simo5: rebased and fixed

mistotebe: Thanks, @bgermann could you update `0029-Load-OpenSSL3-legacy-provider-digestmd5.patch` in Debian accordingly?

bgermann: I won't patch an upstream patch. Please remind me to add this patch once it has arrived in master branch.

PR #690, opened 2021-11-11
Sphinx 4

bgermann:
These are the necessary changes to build the documentation with Sphinx 4.


simo5:
I came up with a similar fix to make distcheck happy. Please apply before 2.1.28

PR #695, opened 2021-11-18
wip: remove java support from 2.1 tree

simo5:
This is just an FYI PR, if someone has difficulty with the 2,1 due to it

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -59,18 +59,13 @@ SAM = sample
 else
 SAM =
 endif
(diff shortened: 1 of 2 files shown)


quanah:
@hyc given that the java subtree is completely unusable, I think this is a happy compromise for 2.1.28 -- If people want it, they can revert this commit, but this will allow people to build on current platforms easily.

hyc (approved):

PR #714, opened 2022-03-08
Fix #700 don't fully qualfy client hostname

### LINKED ISSUE #700: sasl_client_new() should not query DNS for client hostnames
When I use ANONYMOUS authentication to a service (eg AMQP), Cyrus looks up my client's hostname in DNS. It tries to look up my unqualified shortname in DNS. Not only is this unnecessary, but in my particular test environment, this can take several seconds for DNS to respond to both the A and AAAA queries for the unqualified name.

### Steps to reproduce:

1. Set up a CentOS 8 or 9 host.
2. Install cyrus-sasl-devel
3. Ensure the unqualified (short) hostname from the `hostname` command is not present in `/etc/hosts`.
4. Run the following program that simply calls `sasl_client_new()`:
```c
#include <stdio.h>
#include <sasl/sasl.h>

int main() {
   int result;

   result = sasl_client_init(NULL);
   if (result != SASL_OK) {
       printf("sasl failure:\n");
       printf("%s\n", sasl_errstring(result, NULL, NULL));
   }

    sasl_conn_t *conn;

    result = sasl_client_new("amqp",
                             "localhost",
                             NULL, NULL, /* Local and remote IP address
                                            strings (NULL disables mechanisms
                                            which require this info)*/
                             NULL,       /* connection-specific callbacks */
                             0,          /* security flags */
                             &conn);     /* allocated on success */

    if (result != SASL_OK) {
        printf("sasl failure:\n");
        printf("%s\n", sasl_errstring(result, NULL, NULL));
    }
}
```

### Actual results:
The program makes two DNS queries (A and AAAA) for my host's unqualified hostname. The AAAA request takes a while to answer (3.5 seconds) in my environment.

### Expected Results:
There should be zero DNS traffic for my client's hostname.

hyc:
Partially reverts 7ae4341f6226dcb1d4560c665821bc606b8b8ee0

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -398,7 +398,7 @@ int sasl_client_new(const char *service,
 		    sasl_conn_t **pconn)
 {
   int result;
-  char name[MAXFQDNLEN];
+  char name[256];
   sasl_client_conn_t *conn;
   sasl_utils_t *utils;
   sasl_getopt_t *getopt;
@@ -512,9 +512,8 @@ int sasl_client_new(const char *service,
   
   /* get the clientFQDN (serverFQDN was set in _sasl_conn_init) */
   memset(name, 0, sizeof(name));
-  if (get_fqhostname (name, MAXFQDNLEN, 0) != 0) {
-      return (SASL_FAIL);
-  }
+  gethostname(name, sizeof(name));
+  name[sizeof(name)-1] = '\0';
 
   result = _sasl_strdup(name, &conn->clientFQDN, NULL);
 
--- a/plugins/anonymous.c
+++ b/plugins/anonymous.c
@@ -233,7 +233,6 @@ anonymous_client_mech_step(void *conn_context,
 {
     client_context_t *text = (client_context_t *) conn_context;
     size_t userlen;
-    char hostname[256];
     const char *user = NULL;
     int user_result = SASL_OK;
     int result;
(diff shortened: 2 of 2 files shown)


hyc:
Also change the anonymous plugin to actually use the hostname in the given client params, instead of redundantly calling `gethostname()` again.

ktdreyer:
I cherry-picked this change to a branch based on Red Hat's 2.1.27 CentOS 9 package. I built [RPMs](https://ktdreyer.fedorapeople.org/cyrus-sasl-2.1.27-20.kdreyer.el9/) with the change in order to test it.

With this change, I do not see DNS lookups in my test program that calls `sasl_client_new()`.

ktdreyer:
`qualfy` -> `qualify` in the commit message, otherwise LGTM. Thank you.

bgermann:
Why the triple Signed-off-by?

makr:
Maybe out of scope, but I suggest to review this code again and not only the PR. gethostname() can fail, although POSIX doesn't specify any specific error, glibc does. Furthermore glibc documents that if the hostname doesn't fit into the buffer it may not be properly null-terminated. This may then become an issue, with the following strdup().

hyc:
> Why the triple Signed-off-by?

Ah that's my commit template, it must have reinserted after each `commit --amend`.

hyc:
> Maybe out of scope, but I suggest to review this code again and not only the PR. gethostname() can fail, although POSIX doesn't specify any specific error, glibc does. Furthermore glibc documents that if the hostname doesn't fit into the buffer it may not be properly null-terminated. This may then become an issue, with the following strdup().

Good point, the original code in plugins/anonymous.c explicitly used a `char hostname[256]` which is guaranteed to be large enough. It also explicitly NUL-terminates. I've added that here too. The MAXFQDNLEN define is not appropriate for use here since that's only defined to 254.

mistotebe (approved):

PR #716, opened 2022-03-29
Fix #637 don't enable GSSAPI channel bindings by default

### LINKED ISSUE #637: Revert the Channel Binding changes in gssapi.c
The GSSAPI SASL mechanism defined in [RFC 4752 The Kerberos V5 ("GSSAPI") SASL Mechanism ](https://tools.ietf.org/html/rfc4752) says explicitly, that Channel Binding is not supported.

The only SMTP server that support CB is exim (with gsasl, thus for SCRAM and not for GSSAPI), the only IMAP server supporting CB is Cyrus IMAP, no SMTP/IMAP client supports CB, except imtest (part of cyrus imap).

OpenLDAP intends to be compatible with Windows.  Windows apparently does not support GS2-KRB5-PLUS, or GS2-KRB5, but GSSAPI.  Windows can enforce Channel Binding over GSSAPI over LDAP and this is not compatible with RFC 4752.

Cyrus IMAP utilizes unconditionally the Channel Bindings API of Cyrus SASL and therefore can offer both SCRAM-SHA-1 and SCRAM-SHA-1-PLUS.  With the unconditional initialization of Channel Bindings, Cyrus IMAP together with the last change in gssapi.c enforces CB for GSSAPI over IMAP4, POP3, ManageSieve, HTTP/SPNEGO.

I understand OpenLDAP strives to function as Windows AD substitution, and for this non-RFC 4752 compliant GSSAPI SASL method must be used.  OpenLDAP has to achieve this however without the channel bindings in plugins/gssapi.c.  Otherwise Cyrus IMAP cannot offer, provided that it utilizes the API correct, SCRAM-SHA-1, SCRAM-SHA-1-PLUS and GSSAPI at the same time correctly (e.g. for Thunderbird and Evolution which do GSSAPI without Channel Binding, but might one day offer SCRAM with Channel Binding, too).

hyc:

### DIFF
--- a/include/sasl.h
+++ b/include/sasl.h
@@ -304,6 +304,8 @@ typedef unsigned sasl_ssf_t;
  *			   credentials to do so
  * MUTUAL_AUTH          -- require mechanisms which provide mutual
  *			   authentication
+ * NONSTD_CBIND         -- enable channel binding on mechs that aren't
+ *			   supposed to support them but do anyway (e.g. GSSAPI)
  */
 #define SASL_SEC_NOPLAINTEXT      0x0001
 #define SASL_SEC_NOACTIVE         0x0002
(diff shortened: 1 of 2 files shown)


hyc:
We will probably have to tweak libldap to not enable this by default, and leave it up to site admins to configure their sasl secprops to turn it on off needed.

simo5:
LGTM

PR #721, opened 2022-04-14
plugins/digestmd5: Remove debug log "mech free"

### LINKED ISSUE #386: Libsasl spams our logs with "DIGEST-MD5 common mech free" messages
From: [Sjon](sjon@parse.nl)
Bugzilla-Id: 3818
Version: 2.1.x
Owner: [Alexey Melnikov](Alexey.Melnikov@isode.com)

bgermann:
The "DIGEST-MD5 common mech free" debug log message is bothering many users.
It is not really helpful, so drop it.

Fixes #386.

### DIFF
--- a/plugins/digestmd5.c
+++ b/plugins/digestmd5.c
@@ -1762,9 +1762,6 @@ static void digestmd5_common_mech_free(void *glob_context,
     reauth_cache_t *reauth_cache = my_glob_context->reauth;
     size_t n;
 
-    utils->log(utils->conn, SASL_LOG_DEBUG,
-	       "DIGEST-MD5 common mech free");
- 
     /* Prevent anybody else from freeing this as well */
     my_glob_context->reauth = NULL;
 


hyc:
Kind of wondering why the system logger is logging DEBUG messages. Seems to me the real fix is there, and telling it to honor message priority settings.

bgermann:
That is kinda the default systemd behaviour: log everything and filter on viewing. The problem with that is that you end up with a large log in this case. In Debian, the package maintainers ended up with a specific logcheck rule for this log message, which I just removed together with applying this patch: https://salsa.debian.org/debian/cyrus-sasl2/-/commit/b0eeb543ebf7.

See also: https://bugs.debian.org/805310

hyc:
The default log_level in libsasl is SASL_LOG_ERR so with this check https://github.com/cyrusimap/cyrus-sasl/blob/master/lib/common.c#L1488 the message shouldn't even get sent to syslog. If it's getting there, sounds like they've misconfigured something.

hyc:
This is is probably a better approach https://github.com/hyc/cyrus-sasl/tree/issue386

bgermann:
I guess so. Thanks!

hyc (approved):

PR #722, opened 2022-04-14
Fix #386 - honor log_level option on clients too

### LINKED ISSUE #386: Libsasl spams our logs with "DIGEST-MD5 common mech free" messages

hyc:

Supersedes #721

### DIFF
--- a/include/saslplug.h
+++ b/include/saslplug.h
@@ -305,7 +305,7 @@ typedef struct sasl_client_params {
     int (*spare_fptr1)(void);
 
     unsigned int cbindingdisp;
-    int spare_int2;
+    int log_level;
     int spare_int3;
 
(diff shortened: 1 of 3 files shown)


bgermann:
I cannot get this to work. How did you test it?

hyc:
Hm, you're right. The ->conn has been freed and NULL'd already by client_dispose() before it gets to the mech_free() message.

hyc:
I also don't see any other plugins doing anything special in their mech_free() handler. So ok, let's take your patch in addition to this one.

PR #745, opened 2022-08-24
Misc

hyc:
A bunch of tweaks primarily to support building for Windows in MSYS/MinGW.


quanah:
I'd suggest dropping commit [7076530](https://github.com/cyrusimap/cyrus-sasl/pull/745/commits/70765304ad2c99bc96ecef122d7efbb650a08d35) since we're dropping BDB support entirely.

configure.ac
```diff
@@ -1528,15 +1528,21 @@ struct sockaddr_storage {
 #endif
 
 #ifdef WIN32
-#define SASL_ROOT_KEY "SOFTWARE\\Carnegie Mellon\\Project Cyrus\\SASL Library"
-#define SASL_PLUGIN_PATH_ATTR "SearchPath"
-#define SASL_CONF_PATH_ATTR "ConfFile"
+#define SASL_ROOT_KEY "SOFTWARE\\Project Cyrus\\SASL Library"
```
mistotebe, line 1531: Would upgrade instructions be in order? I notice this path wasn't documented anywhere else, should it?

configure.ac
```diff
@@ -1239,8 +1239,8 @@ IPv6_CHECK_FUNC(getaddrinfo, [IPv6_CHECK_FUNC(gai_strerror,
                 sasl_cv_getaddrinfo=no])])
 if test $sasl_cv_getaddrinfo = yes; then
     AC_LIBOBJ(getaddrinfo)
-    GETADDRINFOOBJS="getaddrinfo.o"
-    LTGETADDRINFOOBJS="getaddrinfo.lo"
+    GETADDRINFOOBJS="../lib/getaddrinfo.o"
```
mistotebe, line 1242: Does this avoid the need for f9f28ac09bf35336eb04782258bec8dd5c6767b8 to exist (or make it a deletion)?

lib/server.c
```diff
@@ -90,7 +90,7 @@ static int _sasl_checkpass(sasl_conn_t *conn,
 
 static mech_list_t *mechlist = NULL; /* global var which holds the list */
 
-static sasl_global_callbacks_t global_callbacks;
+static sasl_global_callbacks_t server_global_callbacks;
```
mistotebe, line 93: Client uses `global_callbacks_client`, should this follow the same schema?

include/saslplug.h
```diff
@@ -153,7 +153,10 @@ typedef struct sasl_utils {
     void (*prop_erase)(struct propctx *ctx, const char *name);
     int (*auxprop_store)(sasl_conn_t *conn,
 			 struct propctx *ctx, const char *user);
-
+    /* Registry value functions */
+    int (*get_registry_value)(void *context __attribute__((unused)),
```
mistotebe, line 157: Is this part of the liblsasl ABI? If so, soname (`sasl_version` in `lib/Makefile.am`) would need bumping...

configure.ac
hyc: Probably a good idea. Dunno which doc section that would go in.

configure.ac
hyc: I believe those other files ought to be deleted, yeah. But the purpose of this PR was mainly to import old patches we were already carrying. Other cleanup can be a separate PR.

lib/server.c
hyc: I'd be inclined to go the other way, since global_callbacks_server would make tab completion in gdb slower. As these are static variables, the only reason to rename them is to eliminate ambiguity in the debugger.

include/saslplug.h
hyc: True

lib/server.c
hyc: I'll note that when this patch was written, this variable was a global and not static. So probably the patch was fixing an actual symbol collision before, and that's not an issue in the current code, so we could also just drop this commit.

hyc: Dropped this commit

hyc:
> I'd suggest dropping commit [7076530](https://github.com/cyrusimap/cyrus-sasl/pull/745/commits/70765304ad2c99bc96ecef122d7efbb650a08d35) since we're dropping BDB support entirely.

Yeah, dropped.

configure.ac
hyc: Deleted redundant files

configure.ac
hyc: Done

lib/Makefile.am
```diff
@@ -43,7 +43,7 @@
 # Library version info - here at the top, for sanity
 # See <http://www.gnu.org/software/libtool/manual/libtool.html#Versioning>
 # CURRENT:REVISION:AGE
-sasl_version = 3:0:0
+sasl_version = 4:0:1
```
mistotebe, line 46: I think AGE should be kept at 0 since the layout of `struct sasl_utils` is changed incompatibly and I can't see any other way the library tries to maintain support for plugins linked against `libsasl2.so.3` either.

hyc: Quoting https://www.gnu.org/software/libtool/manual/libtool.html#Versioning :

     If any interfaces have been added, removed, or changed since the last update, increment current, and set revision to 0.
     If any interfaces have been added since the last public release, then increment age.

mistotebe: My reading is that the `struct sasl_utils` part of the interface is "changed" (a plugin populating or accessing that struct will not work with new ABI) so we can't support v3 and v4 simultaneously (which we claim when setting AGE to 1).

hyc: An old plugin will work with a new library. The added fields just occupy the "spare" fields in the old structure so old plugins would ignore them.

mistotebe: I don't understand, it looks like the structure has just grown by 2 pointers in the middle (shifting everything other member down a little), I can't square that with your suggestion.

mistotebe: Then those spare pointers should be considered replaced by this, not added, so as not to affect the struct's size?

hyc: Irrelevant. The struct is allocated by libsasl, plugins don't know or care about its size.

mistotebe: To sum up the above and an out-of-band conversation:

`struct sasl_utils` is meant to be set up by liblsasl and considered read-only by plugins, so anything past its end is not considered part of the ABI. Adding these, we have to tell new users that the members are available (moving our soname to liblsasl2.so.4) but we haven't changed the meaning of anything in the v3 ABI and anyone linked against liblsasl2.so.3 should work just as it did before (and we can bump AGE to 1 to say we support both interfaces).

Given we can do the same next time something like this is needed, it turns out those extra pointers were never necessary and being unused, we can and will drop them. This might also have the side effect of helping automated ABI checkers understand the intent of such changes.

mistotebe (approved):

SyntevoAlex:
Commit https://github.com/cyrusimap/cyrus-sasl/commit/79827fbe
broke Windows build because it removed `return_value` that is still present in code
https://github.com/cyrusimap/cyrus-sasl/blob/38dc8b804380330f21153ff3d016109faeec1611/lib/common.c#L2692
https://github.com/cyrusimap/cyrus-sasl/blob/38dc8b804380330f21153ff3d016109faeec1611/lib/common.c#L2696

SyntevoAlex:
Commit https://github.com/cyrusimap/cyrus-sasl/commit/b2a80d38
broke Windows build when using Visual C++ build tools because `SASL_DB_PATH_ATTR`
https://github.com/cyrusimap/cyrus-sasl/blob/38dc8b804380330f21153ff3d016109faeec1611/sasldb/allockey.c#L283
is only defined in `configure.ac` which is not used through `NTMakeFile`

Issue #746, opened 2022-08-29
runtests.py ：The return value of the  gssapi_channel_binding_mismatch_test() function is incorrect.


winner658:
in tests/runtests.py file，when the gssapi_channel_binding_mismatch_test() runs normally and is bound successfully, the gssapi_channel_binding_mismatch_test() returns an incorrect result value instead of a correct result value. When an exception occurs, the gssapi_channel_binding_mismatch_test() returns a correct result value. As a result, the following error message is displayed after all tests are passed in time:
![image](https://user-images.githubusercontent.com/14918864/187192919-fc7798da-88e9-407f-bb83-39fa22ef1f48.png)
 
in other function: the return value is 0 for success and 1 for exception.
![image](https://user-images.githubusercontent.com/14918864/187192975-a7aff770-0831-4058-bd96-cd7e7a4f593b.png)

hyc:
I don't believe your assessment is correct. The test clearly says "This test should fail" which means a successful bind is a failure, not a success.

PR #747, opened 2022-08-29
Update runtests.py: Fixed an issue where the return value of the function is incorrect.

winner658:
Fixed an issue where the return value of the function is incorrect.[https://github.com/cyrusimap/cyrus-sasl/issues/746](url)
When the gssapi_channel_binding_mismatch_test function is bound successfully, 0 is returned. When an exception occurs, 1 is returned. The function is consistent with other functions of the same type.

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -213,9 +213,10 @@ def gssapi_channel_binding_mismatch_test(kenv):
                 result = "PASS"
             raise Exception("CLI ({}): {} --> SRV ({}): {}".format(
                 cli.returncode, cli_err, srv.returncode, srv_err))
+        return 0
     except Exception as e:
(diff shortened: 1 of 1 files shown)


winner658:
The local test runs normally. Why is the failure displayed here? Where can I view the log?

bgermann:
The Signed-off-by has to be in the commit message, not in the PR description.

mistotebe:
Except this test is making sure that the exchange actually fails (in which case it returns `0` - success), so the right way around would be to fix the messages emitted. And probably make sure that the result of `plain_tests()` contributes to `err`.

Neustradamus:
@winner658: Any progress on this PR 3 years after?

Have you seen @quanah and @mistotebe comments?

PR #750, opened 2022-11-20
Do not add unneeded `-lcrypt` in `plain`, `login` plugins

SyntevoAlex:
The problem is that on newer Linux, for example Arch Linux 2022-09, the deprecated library is missing. This causes plugins to fail to load.

It was last needed more than 20 years ago, and was removed in commit
  74135b68 by Larry Greenfield, 1999-06-23 00:16:37
  latest stuff

Later, `#include <crypt.h>` was also removed in commit
  f58e0357 by Larry Greenfield, 2000-02-27 23:46:15
  removed superfluous includes, brought in line with new checkpass()

### DIFF
--- a/m4/plain.m4
+++ b/m4/plain.m4
@@ -1,21 +1,11 @@
 dnl Check for PLAIN (and therefore crypt)
 
 AC_DEFUN([SASL_PLAIN_CHK],[
-AC_REQUIRE([SASL2_CRYPT_CHK])
 
 dnl PLAIN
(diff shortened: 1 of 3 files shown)


SyntevoAlex:
Hmm, I didn't intend to. Probably some kind of accident. Thanks for noticing!

quanah:
Great, thanks!

quanah:
@hyc wonder if we can remove libcrypt entirely from cyrus-sasl?

hyc (approved):

SyntevoAlex:
Perfect, thanks for taking care of my PR!

PR #753, opened 2022-12-25
sasl: fix sqlite3 plugin value inserts

SyntevoAlex:
Consider the following configuration:

sqlite3 database `sqlite.db`:
```
CREATE TABLE data(username TEXT, realm TEXT, property TEXT, value TEXT)
```

`saslpasswd.conf` file
```
pwcheck_method: auxprop
mech_list: CRAM-MD5
auxprop_plugin: sql
sql_engine: sqlite3
sql_database: sqlite.db
sql_select: SELECT value FROM data WHERE username = '%u' AND realm = '%r' AND property = '%p'
sql_insert: INSERT INTO data (username, realm, property, value) VALUES ('%u', '%r', '%p', '%v')
sql_update: UPDATE data SET value='%v' WHERE username = '%u' AND realm = '%r' AND property = '%p'
```

Now, when using `saslpasswd2 -u TestRealm -c TestUser` it pretended to succeed, but in fact did nothing.
Please refer to commit message for the explanation.

### DIFF
--- a/plugins/sql.c
+++ b/plugins/sql.c
@@ -524,18 +524,41 @@ static int _sqlite3_escape_str(char *to, const char *from)
     return 0;
 }
 
-static int sqlite3_my_callback(void *pArg, int argc __attribute__((unused)),
-			      char **argv,
-			      char **columnNames __attribute__((unused)))
-{
-    char **result = (char**)pArg;
+typedef struct sql_callback_data {
+    char *value;
+    size_t size;
+    size_t *value_len;
+    size_t num_columns;
+    size_t num_values;
+} sql_callback_data_t;
 
-    if (argv == NULL) {
-	*result = NULL;				/* no record */
-    } else if (argv[0] == NULL) {
-	*result = strdup(SQL_NULL_VALUE);	/* NULL IS SQL_NULL_VALUE */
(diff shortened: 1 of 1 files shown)


SyntevoAlex:
Could be related: https://lists.andrew.cmu.edu/pipermail/cyrus-sasl/2020-July/003283.html

plugins/sql.c
```diff
@@ -524,18 +524,41 @@ static int _sqlite3_escape_str(char *to, const char *from)
...
+	value = SQL_NULL_VALUE;		/* NULL IS SQL_NULL_VALUE */
+	data->num_values++;
     } else {
-	*result = strdup(argv[0]);
+	value = column_values[0];
+	data->num_values++;
+    }
+
+    /* now get the result set value and value_len */
+    /* we only fetch one because we don't care about the rest */
```
hyc, line 555: Getting more than one result should be a failure, certainly it is a security vulnerability to have ambiguous credentials. If an attacker can insert their own creds in front of the valid ones, etc...

Neustradamus:
@SyntevoAlex: Have you seen the @hyc comment?

SyntevoAlex:
Sorry, I will no longer work on this PR, because I no longer work with the employer. It can be closed now.

PR #763, opened 2023-05-17
Support multiple service principals in a single GSSAPI service

cgaspar-deshaw:
With the current code, it's impossible for a single service to accept multiple service principals. This prevents a single server from being known by multiple FQDNs (e.g. the host FQDN and an anycast FQDN). The code also makes assumptions about how the service principal name should be constructed. This PR addresses both issues by adding support for 2 new options:

accept_any_principal: do not specify the service principal name
service_principal: specify the desired service principal name instead of generating it

I'd argue that best practice would be to make accept_any_principal the norm and delete a lot of code, but I didn't want to change existing behaviour.

I also added client support for the existing keytab option.

### DIFF
--- a/plugins/gssapi.c
+++ b/plugins/gssapi.c
@@ -1114,59 +1114,80 @@ gssapi_server_mech_authneg(context_t *text,
     gss_OID mech_type;
     gss_channel_bindings_t bindings = GSS_C_NO_CHANNEL_BINDINGS;
     struct gss_channel_bindings_struct cb = {0};
+    unsigned int rl; /* Required storage by getopt() */
+    const char *service_principal_opt = NULL; /* The configured service principal name */
+    unsigned char accept_any = 0;
+    const sasl_utils_t *utils = text->utils;
+#ifdef GSS_USE_LOAD_FROM
+    char* keytab = NULL; /* The configured credentials cache file path */
+
+    utils->getopt(utils->getopt_context, "GSSAPI", "keytab", (const char**)&keytab, &rl);
+#endif
 
     input_token = &real_input_token;
     output_token = &real_output_token;
     output_token->value = NULL; output_token->length = 0;
     input_token->value = NULL; input_token->length = 0;
-    
+
     if (text->server_name == GSS_C_NO_NAME) { /* only once */
-	if (params->serverFQDN == NULL
-	    || strlen(params->serverFQDN) == 0) {
-	    SETERROR(text->utils, "GSSAPI Failure: no serverFQDN");
-	    sasl_gss_free_context_contents(text);
-	    return SASL_FAIL;
+	utils->getopt(utils->getopt_context,
+		      "GSSAPI", "service_principal", (const char**) &service_principal_opt, &rl);
+	if (service_principal_opt &&
+	    strlen(service_principal_opt) == 1 &&
+	    service_principal_opt[0] == '*') {
+	    accept_any = 1;
 	}
-	name_token.length = strlen(params->service) + 1 + strlen(params->serverFQDN);
-	name_token.value = (char *)params->utils->malloc((name_token.length + 1) * sizeof(char));
-	if (name_token.value == NULL) {
-	    MEMERROR(text->utils);
-	    sasl_gss_free_context_contents(text);
-	    return SASL_NOMEM;
-	}
-	sprintf(name_token.value,"%s@%s", params->service, params->serverFQDN);
+	if (!accept_any && text->server_name == GSS_C_NO_NAME) {
+	    if (service_principal_opt) {
+		name_token.length = strlen(service_principal_opt);
+	    } else {
+		if (params->serverFQDN == NULL
+		    || strlen(params->serverFQDN) == 0) {
+		    SETERROR(text->utils, "GSSAPI Failure: no serverFQDN");
+		    sasl_gss_free_context_contents(text);
+		    return SASL_FAIL;
+		}
+		name_token.length = strlen(params->service) + 1 + strlen(params->serverFQDN);
+	    }
 
-	GSS_LOCK_MUTEX_CTX(params->utils, text);
-	maj_stat = gss_import_name (&min_stat,
-				    &name_token,
-				    GSS_C_NT_HOSTBASED_SERVICE,
-				    &text->server_name);
-	GSS_UNLOCK_MUTEX_CTX(params->utils, text);
+	    name_token.value = (char *)params->utils->malloc((name_token.length + 1) * sizeof(char));
+	    if (name_token.value == NULL) {
+		MEMERROR(text->utils);
+		sasl_gss_free_context_contents(text);
+		return SASL_NOMEM;
+	    }
 
-	params->utils->free(name_token.value);
-	name_token.value = NULL;
+	    if (service_principal_opt) {
+		strncpy(name_token.value, service_principal_opt, name_token.length + 1);
+	    } else {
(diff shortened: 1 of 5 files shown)


simo5:
I think this is unnecessary, you should be able to do what you want by setting:
ignore_acceptor_hostname in krb5.conf's libdefault section.

cgaspar-deshaw:
The issue with krb5.conf is that this is a GSSAPI app, so there's no way to set it except globally or via an environment variable, so it applies (minimally) to an entire process. If a user wants different behaviour on different endpoints, krb5.conf is not an option.

Also "you can work around bad code by doing this hack, so don't fix the code" is an ... interesting ... line of reasoning.

simo5:
I am not against code that deals with this withing cyrus-sasl, but there is a mechanism that works, and is part of libgssapi. It does require a change in system config, and there may be odd cases where you may want slightly different behaviors between different parts of an application such that setting an application specific krb5.conf is not I guess?
It just sounds very niche.

plugins/gssapi.c
```diff
@@ -1114,32 +1114,63 @@ gssapi_server_mech_authneg(context_t *text,
...
+	    strncpy(name_token.value, service_principal_opt, name_token.length + 1);
+	} else {
+	    sprintf(name_token.value,"%s@%s", params->service, params->serverFQDN);
+	}
 
 	GSS_LOCK_MUTEX_CTX(params->utils, text);
 	maj_stat = gss_import_name (&min_stat,
 				    &name_token,
-				    GSS_C_NT_HOSTBASED_SERVICE,
+				    service_principal_opt ? GSS_KRB5_NT_PRINCIPAL_NAME : GSS_C_NT_HOSTBASED_SERVICE,
```
simo5, line 1173: Why mixing types here?
The service principal name should be a GSSAPI name specified as svc@name.

If you want to support *kerberos* principal names, you should name the parameter differently, but I hardly see a reason to do that.

cgaspar-deshaw:
This PR also addresses other code issues as well - please take a look at the code. The biggest one is not trying to load credentials when not needed, which caused thread deadlocks in OpenLDAP when used with Heimdal and a KCM ccache (which is really a Heimdal bug, but this avoids it). I (hopefully) did so in a way that does not impact backwards compatibility.

plugins/gssapi.c
```diff
@@ -1210,6 +1235,29 @@ gssapi_server_mech_authneg(context_t *text,
 	    server_creds = text->server_creds;
 	}
     }
+#ifdef GSS_USE_LOAD_FROM
+    else if (accept_any && keytab != NULL) {
```
simo5, line 1239: I believe you should do this only if server_creds == GSS_C_NO_CREDENTIAL like the above code, and may make sense to rework the code to just sahre the above code path.
the only difference is text->server_name vs GSS_C_NO_NAME, just use a supporting variable and collapse the code into one code path ?

plugins/gssapi.c
```diff
@@ -2014,6 +2062,12 @@ static int gssapi_client_mech_step(void *conn_context,
     gss_cred_id_t client_creds = (gss_cred_id_t)params->gss_creds;
     gss_channel_bindings_t bindings = GSS_C_NO_CHANNEL_BINDINGS;
     struct gss_channel_bindings_struct cb = {0};
+    const char *service_principal_opt = NULL; /* The configured service principal name */
+#ifdef GSS_USE_LOAD_FROM
+    char* keytab = NULL; /* The configured credentials cache file path */
```
simo5, line 2067: The comment is wrong, a keytab is not a credential cache.

I am also not sure you should use "keytab" on the client side. GSSAPI distinguishes between "keytab" and "client_keytab" for good reasons.

simo5 (changes requested):

quanah:
> This PR also addresses other code issues as well - please take a look at the code. The biggest one is not trying to load credentials when not needed, which caused thread deadlocks in OpenLDAP when used with Heimdal and a KCM ccache (which is really a Heimdal bug, but this avoids it). I (hopefully) did so in a way that does not impact backwards compatibility.

I would advise opening separate PRs for bug fixes.  I'd also suggest [opening a bug](https://github.com/heimdal/heimdal/issues) with Heimdal if you've found an issue with their ccache.

plugins/gssapi.c
```diff
@@ -1114,32 +1114,63 @@ gssapi_server_mech_authneg(context_t *text,
     gss_OID mech_type;
     gss_channel_bindings_t bindings = GSS_C_NO_CHANNEL_BINDINGS;
     struct gss_channel_bindings_struct cb = {0};
+    unsigned int rl; /* Required storage by getopt() */
+    const char *service_principal_opt = NULL; /* The configured service principal name */
+    const char *any_principal = NULL; /* Do we accept any principal in the server keytab? */
```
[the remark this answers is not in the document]
cgaspar-deshaw: Changed to overload service_principal_opt instead.

plugins/gssapi.c
cgaspar-deshaw: GSS_C_NT_HOSTBASED_SERVICE can't express all valid KRB5 principal names, but I don't care enough to keep it in. Switched to always using GSS_C_NT_HOSTBASED_SERVICE.

plugins/gssapi.c
cgaspar-deshaw: Re-worked the code in question - not sure if it's an improvement, but does remove the near-duplicate code.

plugins/gssapi.c
cgaspar-deshaw: Removed all client changes. I have no immediate need for them.

plugins/gssapi.c
```diff
@@ -2149,10 +2213,28 @@ static int gssapi_client_mech_step(void *conn_context,
...
-                maj_stat = gss_acquire_cred(&min_stat, GSS_C_NO_NAME,
-                                            GSS_C_INDEFINITE, desired_mechs,
-                                            GSS_C_INITIATE,
-                                            &text->client_creds, NULL, NULL);
+#ifdef GSS_USE_LOAD_FROM
+                if (keytab != NULL) {
+                    gss_key_value_element_desc keytab_element = {.key = "keytab", .value = keytab};
+                    gss_key_value_element_desc elements[1] = {keytab_element};
+                    gss_key_value_set_desc cred_store = {.elements = elements, .count = 1};
+
```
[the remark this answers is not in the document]
cgaspar-deshaw: Removed all client changes. I have no immediate need for them.

simo5:
Looks much better now

plugins/gssapi.c
```diff
@@ -1114,59 +1114,80 @@ gssapi_server_mech_authneg(context_t *text,
...
 
     input_token = &real_input_token;
     output_token = &real_output_token;
     output_token->value = NULL; output_token->length = 0;
     input_token->value = NULL; input_token->length = 0;
-    
+
     if (text->server_name == GSS_C_NO_NAME) { /* only once */
-	if (params->serverFQDN == NULL
-	    || strlen(params->serverFQDN) == 0) {
```
hyc, line 1125: I acknowledge that this code was there already but if you're going to be making fixes in here:

Calling strlen() multiple times on the same string is just gross. Would prefer if it was only called once, with the return value saved if it's going to be referenced more than once. Also, checking `strlen(foo) == 0` is a stupid waste of a function call, should just be checking `*foo == '\0'` or just `!*foo`.

hyc: Note that cleanup of the sort I suggest would be better in a separate commit. The current commit can be left as-is.

PR #765, opened 2023-05-21
Update canonusr.c

rowingdude:
I made the following changes to the code:

    Improved variable naming and added comments for better code understanding.
    Reorganized the code and separated functions into logical sections.
    Refactored the _sasl_canon_user_lookup function for better readability.
    Improved error handling and added appropriate error messages.
    Removed unnecessary headers and unused variables.
    Made minor code formatting improvements for consistency.
    Ensured that memory is properly allocated and freed where necessary.
    Made the code more modular and structured by separating functions into individual files.

### DIFF
--- a/lib/canonusr.c
+++ b/lib/canonusr.c
@@ -1,8 +1,91 @@
 /* canonusr.c - user canonicalization support
  * Rob Siemborski
  */
+
+#include <config.h>
+#include <sasl.h>
+#include <string.h>
+#include <ctype.h>
+#include <prop.h>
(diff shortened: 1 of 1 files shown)


GuidoKiener:
It's not a good idea to mix reformatting, fixes, and reorganizing code in one commit. Nobody wants to review such a commit, and readers cannot easy see what you have fixed or changed.

Indeed the format of cyrus sasl code is not pretty, but it should not be changed anymore, since it would be a nightmare if someone starts to reformat the complete code.

PR #767, opened 2023-05-28
MD5 clean-up

bgermann:
Replace the internal MD5 implementation with OpenSSL.
This is only compile-tested.


simo5:
If you are moving to OpenSSL in 2023 you should really use the EVP interface, the old MD5 interface is deprecated and can be compiled out (and at some point OpenSSL will compile it out by default).

bgermann:
Can you copy arbitrary values to the MD5 state as used in the _sasl_hmac_md5_precalc function?

simo5:
> Can you copy arbitrary values to the MD5 state with EVP as used in the _sasl_hmac_md5_precalc function?

I do not think so ... if you need custom MD5 manipulation you may want to retain custom code then, until we get rid of md5 once and for all ...

bgermann:
I have updated at least the saslauthd cache to use EVP.

lib/md5.c
```diff
@@ -103,8 +88,8 @@ void _sasl_hmac_md5_init(HMAC_MD5_CTX *hmac,
    */
 
   /* start out by storing key in pads */
-  MD5_memset((POINTER)k_ipad, '\0', sizeof k_ipad);
-  MD5_memset((POINTER)k_opad, '\0', sizeof k_opad);
+  memset(k_ipad, '\0', sizeof k_ipad);
+  memset(k_opad, '\0', sizeof k_opad);
```
simo5, line 84: memset will be optimized out in many compilers, you may want consider using explicit_bzero() where supported (glibc 2.25 and later).

simo5:
Looks like a great improvement and clean up to me.
@quanah I think this is worth adding to the next release.

lib/md5.c
simo5: I understand, but the problem is that you are introducing a regression here.

All modern compilers have code elision detection that is applied to memset because it is a well known function, so it has much higher chances of being elided than a custom function like MD5_memset.

So while I think we should remove MD5_memset I also think we should try to use explict_bzero which is guaranteed to not be elided.

The whole point of the memsets at the end of the function is to make sure keys are not left in memory, but code elision would leave them there.

bgermann: I have come up with the portable solution to use OPENSSL_cleanse.

simo5: LGTM

hyc (approved):

Issue #751, opened 2022-12-08
Remove support for RC4


hol430:
Hi all, I have stumbled across a segfault which I think may be caused by libsasl 2.1.28 + openssl3 when using rc4 algorithm.

I encountered the problem when attempting to `svn commit`. Our svn server is using sasl auth with rc4 algorithm. I've created [this script](https://github.com/cyrusimap/cyrus-sasl/files/10182202/sasl-repro-sh.txt) which reproduces the problem by starting a local svn server with sasl auth, and attempting to authenticate to the server. When I encountered the problem in the wild, I received [this stack trace](https://github.com/cyrusimap/cyrus-sasl/files/10182211/svn.trace.txt). When the script is run in an appropriate environment (e.g. on my laptop), it generates a [slightly different](https://github.com/cyrusimap/cyrus-sasl/files/10182219/enc_rc4.trace.txt) trace, as the error is now occurring server-side.

I am not sure if the problem is in sasl or openssl code, but I currently suspect sasl. The problem started occurring for me when I upgraded libsasl from 2.1.28-1 (which I believe was built against openssl 1.1) to libsasl 2.1.28-3 (built against openssl3). If I downgrade to older libsasl (ie 2.1.28-1), the problem disappears.

Package Versions:

openssl 3.0.7
libsasl 2.1.28
svn version 1.14.2

Let me know if I missed anything...I'm happy to locally run whatever tests may be helpful.

hol430:
Thanks for the advice; applying the following patches fixed the problem:

887dbc0435056ec58ee48c4d803f110ade1e4c39
dfaa62392e7caecc6ecf0097b4d73738ec4fc0a8
c2bd3afbca57f176d8c650670ce371444bb7fcc0
54f69880fa92bb0d0cf4d55bab0914822a873d8d
725df6cdadc11cf1bbbfa3a57982ec19624c6fbe

Moving away from RC4 does seem like the sensible solution, however my options are limited somewhat by the server configuration.

hyc:
It seems moving off of RC4 should be made a high priority now.

https://www.blackhat.com/eu-22/briefings/schedule/index.html#breaking-kerberos-rc-cipher-and-spoofing-windows-pacs-29181

quanah:
Going to re-open for removal of RC4 support.

nieder:
This also affects tests for OpenLDAP-2.6.4.
[...]

PR #772, opened 2023-07-04
digestmd5: Drop RC4 support

bgermann:
RC4 is an optional cipher in DIGEST-MD5.
It is considered weak and should not be used anymore. Drop it.

Fixes: #751


Neustradamus:
To follow

hyc (approved):

PR #774, opened 2023-07-07
srp: Replace HMAC with EVP_MAC implementation

bgermann:
The openssl/hmac.h interface is deprecated and might not be available in modern OpenSSL configurations. Replace the HMAC computation in SRP with the equivalent EVP_MAC* calls.

### DIFF
--- a/plugins/srp.c
+++ b/plugins/srp.c
@@ -86,7 +86,6 @@ typedef unsigned short uint32;
 
 /* for digest and cipher support */
 #include <openssl/evp.h>
-#include <openssl/hmac.h>
 #include <openssl/md5.h>
 
 /* for legacy libcrypto support */
@@ -168,7 +167,7 @@ typedef struct layer_option_s {
     unsigned enabled;		/* enabled?  determined at run-time */
     unsigned bit;		/* unique bit in bitmask */
     sasl_ssf_t ssf;		/* ssf of layer */
-    const char *evp_name;	/* name used for lookup in EVP table */
+    char *evp_name;		/* name used for lookup in EVP table */
 } layer_option_t;
 
 static layer_option_t digest_options[] = {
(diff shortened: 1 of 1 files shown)


hyc:
I wonder if we need to make sure a valid MD was found at https://github.com/cyrusimap/cyrus-sasl/blob/7fc4d5bb673a81bb3cb93d4854b8a78e008c6f98/plugins/srp.c#L1393

plugins/srp.c
```diff
@@ -1445,14 +1445,15 @@ static int LayerInit(srp_options_t *opts, context_t *text,
...
 
 	/* Initialize the HMACs */
-	text->hmac_md = EVP_get_digestbyname(opt->evp_name);
-        text->hmac_send_ctx = HMAC_CTX_new();
-	HMAC_Init_ex(text->hmac_send_ctx, text->K, text->Klen, text->hmac_md, NULL);
-        text->hmac_recv_ctx = HMAC_CTX_new();
-	HMAC_Init_ex(text->hmac_recv_ctx, text->K, text->Klen, text->hmac_md, NULL);
+	params[0] = OSSL_PARAM_construct_utf8_string("digest", opt->evp_name, 0);
+	params[1] = OSSL_PARAM_construct_end();
+	text->hmac_send_ctx = EVP_MAC_CTX_new(EVP_MAC_fetch(NULL, "HMAC", NULL));
```
hyc, line 1450: Would prefer the EVP_MAC_fetch() was only called once, and the return value reused for both CTX_new() calls.

bgermann:
The EVP_get_digestbyname is not changed by this commit.

hyc:
> The EVP_get_digestbyname is not changed by this commit.

Right. It just stood out as I was reviewing the rest of the file. If you're going to be making changes to the file anyway, might be worth it to tweak this too, unless you know that the digest name coming in over the wire is always going to be valid.

plugins/srp.c
bgermann: Applied.

quanah:
Is this ready for review again?

bgermann:
Yes, I'd rather spare changing the EVP_get_digestbyname call.

hyc (approved):

PR #777, opened 2023-07-17
Revert "Add a HMAC wrapper for deprecated function"

bgermann:
This reverts one commit from #653.

The HMAC() function was never deprecated as opposed to HMAC_Init() and other related functions.
The alternative HMAC implementation is unnecessary.


bgermann:
cc @simo5

hyc (approved):

Issue #790, opened 2023-08-04
[Windows] Test suite segmentation fault


sfhacker:
Ciao.

OS   : Windows
ARCH : x86_64

SASL Version: master branch
SASL DB       : LMDB / Static

When running testsuite file, an exception is thrown:

![image](https://github.com/cyrusimap/cyrus-sasl/assets/25817529/5f37bddb-b47e-446e-adf6-5ab267703832)

This project seems to be in bad shape. We have also faced some compilation issues regarding this file: kdf_hkdf_sha512 among other issues.

Is there any CI server in place? Perhaps that can help.

Thanks in advance.

quanah:
Master is under significant churn in preparation for cyrus-sasl 2.2.  I would note that the test suite issues are not limited to master, see #758

alls0rts:
I have run into this as well, it's a simple fix.  I shall create a pull request.

PR #792, opened 2023-08-10
testsuite SEGV due to index error in testseclayer()

alls0rts:
Running testsuite on Solaris seeing a SEGV for every mech.  Its an indexing issue, looks to be the same issue reported for windows by  #790


quanah:
Please sign off the commit as required to meet the DCO.  Thanks!

quanah:
Also be handy if you updated the commit message to start with Fixes #790
Thanks!

PR #793, opened 2023-08-13
Fix windows build

SyntevoAlex:
This doesn't fix all problems because I ran out of steam.

I managed to build successfully by quick&dirty supressing 5 other problems.
2 of which are from https://github.com/cyrusimap/cyrus-sasl/pull/745#issuecomment-1676143030


Neustradamus:
@hyc: Have you seen this PR?

hyc:
Yes. I've looked it over but it's modifying the MSVC Makefiles and I don't use MSVC.

hyc (approved):
Not tested but looks fine to me.

PR #796, opened 2023-08-18
retire custom manpage generator

elliefm:
For some unknown historical reason, our documentation build uses a custom manpage generator, rather than the stock one provided by Sphinx.  It keeps getting broken by incompatible new Sphinx versions.

This PR switches the docs build to use the stock manpage generator, and then removes the parts we're no longer using.

Equivalent changes were already landed for cyrus-imapd as https://github.com/cyrusimap/cyrus-imapd/pull/4561

**Reviewers:** Two paths to test, each starting from a clean repository, and assuming you have all the prereqs for building the docs:
1. `autoreconf -i -s && ./configure && make man` => lots of warnings (that were already there, and which I'm not trying to address), but should succeed.  Produces man pages in `man/`, which can be viewed by e.g. `man man/sasl_callbacks.3`.  Check that they're not empty files and are formatted okay.
2. `make -C docsrc man` => different output, same warnings, should succeed.  Produces man pages in `docsrc/build/man`, verify as above.


elliefm:
Oh right, I forgot you guys were using those, my bad!

elliefm:
Rebased with signoffs and forced-pushed :)

PR #803, opened 2023-09-05
Init with threads

mistotebe:
Add `--with-threads=<implementation>` configure option that makes `sasl_client_init`/`sasl_server_init` thread safe. Some code is borrowed from OpenLDAP, @hyc can you review this is OK copyright-wise? Some of it (e.g. Windows side of `_sasl_mutex_init`, probably the whole NT infra) is yours.

### DIFF
--- a/m4/ax_pthread.m4
+++ b/m4/ax_pthread.m4
@@ -1,0 +1,522 @@
+# ===========================================================================
+#        https://www.gnu.org/software/autoconf-archive/ax_pthread.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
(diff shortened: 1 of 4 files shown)


m4/ax_pthread.m4
```diff
@@ -0,0 +1,522 @@
+# ===========================================================================
```
hyc, line 1: I doubt it's worth adding this at all. Just use `$(CC) -pthread` and be done with it. Any older compiler isn't worth supporting, any older OS probably isn't in use any more.

mistotebe: It was easier for me to pull the "official" test than write and debug a new one.

mistotebe:
Assuming that you're happy with the copyright side of things and dropping the `Draft` state of this PR.

PR #805, opened 2023-09-12
saslauthd: add -f (foreground) option

sp-andwei:
add -f option to disable daemonizing without activating verbose output at the same time

Useful for service supervision schemes like runit, that rely on the supervised process to run in foreground. Coupling foreground + verbose in this case is not really useful. This does not change the behavior of the existing options.


quanah:
@hyc please review and lmk if you think 2.2 is a good target, thanks!

Issue #811, opened 2023-10-15
SCRAM: Client does not check low iteration counter


GuidoKiener:
**Problem:** Using mechanism SCRAM, the client does not abort authentication when the given iteration counter is lower than 4096.

See https://github.com/cyrusimap/cyrus-sasl/blob/cyrus-sasl-2.1.28/plugins/scram.c#L2457

A hostile server can send a small iteration counter (e.g. 1) and forces the client to send a ClientProof that is calculated with lowest computation time. Thus the hostile server can faster recover the client's password with an offline dictionary or brute-force attack.

Neustradamus:
To follow this ticket

quanah:
fixed alredy

PR #813, opened 2023-10-15
SCRAM: Client does not check low iteration counter #811

GuidoKiener:
Using mechanism SCRAM, a client does not abort authentication when the given iteration counter is lower than 4096.

A hostile server can send a small iteration counter (e.g. 1) and forces the client to send a ClientProof that is calculated with lowest computation time. Thus the hostile server can recover the client's password faster with an offline dictionary or brute-force attack.

This fix compares the iteration counter with the recommended minimum of 4096 and aborts the authentication if the server violates the recommended minimum.

### DIFF
--- a/plugins/scram.c
+++ b/plugins/scram.c
@@ -2455,8 +2455,11 @@ scram_client_mech_step2(client_context_t *text,
     }
 
     if (text->iteration_count < MIN_ITERATION_COUNTER) {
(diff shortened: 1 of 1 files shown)


mistotebe:
PRs should be targeting master, please adjust.

GuidoKiener:
> PRs should be targeting master, please adjust.

Thanks. It's now reopened in #815.

PR #816, opened 2023-10-27
More Win32 build fixes

hyc:
Some missing #includes, minor stuff.

Reminder: you must run configure with `--without-saslauthd --without-authdaemond` since there's no Windows IPC support in either of those.

### DIFF
--- a/lib/common.c
+++ b/lib/common.c
@@ -72,6 +72,7 @@
 #define _WIN32_WINNT 0x0400
 #include <windows.h>
 #include <process.h>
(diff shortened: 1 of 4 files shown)


hyc:
These are relevant when building under MSYS2. Also need to omit the tests/ directory from the build, or patch that to use Winsock. There are a few other build glitches left to resolve, like sample/ needing to add libs/libobj.la to the link for client and server.

mistotebe (approved):
LGTM

PR #817, opened 2023-10-28
Fix docs build

bgermann:
The docs build fails with current sphinx versions with several errors.
Fix them all.


rjbs:
I haven't looked at this yet, but my assumption is that it is no longer needed after changes merged today.

quanah:
fixed by #866

PR #819, opened 2023-12-12
Fix heap corruption

GuidoKiener:
Calculation of resultlen is wrong. E.g. if server allows only one mechanism SCRAM-SHA-256, the expected string for the mechlist_buf is "SCRAM-SHA-256-PLUS SCRAM-SHA-256" with a required size of 33 bytes and not 32 bytes.
Note that (strlen(mysep) * (s_conn->mech_length - 1) * 2) = 0 when s_conn->mech_length = 1.


Neustradamus:
@GuidoKiener: Good!

cc: @aamelnikov.

hyc (approved):

Issue #823, opened 2024-01-02
Server should support alternate types of Channel Bindings


GuidoKiener:
**Problem**

According to RFC 5929 there are different channel binding types (e.g. "tls-unique", "tls-exporter", "tls-end-point-server") for TLS connections that can be used with SASL mechanisms like SCRAM-SHA-*-PLUS or GS2-KRB5-PLUS. After a TLS connection is established a server can currently set only one binding data with `sasl_setprop(conn, SASL_CHANNEL_BINDING, &cb);`.

Depending on protocol agreements a client can select e.g. between:
- "tls-unique" and "tls-end-point-server" when using TLS 1.2
- "tls-exporter" and "tls-end-point-server" when using TLS 1.3

If a server does not support the desired channel binding type of the client then the authentication will fail or the client is obliged to try another channel binding type or even another worse SASL mechanism.
Therefore it's useful to add a new property or callback function that allows the server to select the desired channel binding data.

**Proposal: New Callback Function**

My preferred solution is adding a new callback function:

```C
/* callback to change channel binding type (e.g. "tls-end-point-server")
 * of the server.  The callback is used by plugins like SCRAM-SHA-*-PLUS or
 * GS2-KRB5-PLUS when the desired binding type of the client does not match
 * the binding type set with property SASL_CHANNEL_BINDING. A server should
 * check the requested 'cbindingname' of the 'plugin' and overwrite the
 * channel binding data of property SASL_CHANNEL_BINDING within the callback
 * function.
 *  plugin        -- name of plugin
 *  cbindingname  -- name of desired channel binding type
 */
typedef int sasl_server_channel_binding_t(sasl_conn_t* conn,
					  void* context,
					  const char* plugin,
					  const char* cbindingname);
#define SASL_CB_SERVER_CHANNEL_BINDING (0x8006)
```

This solution is flexibel and a server application can react to specific plugins and binding types.

**Alternative Proposal: New property SASL_CHANNEL_BINDING_2**

Another way is to add a new property SASL_CHANNEL_BINDING_2 that allows the server to set an alternative channel binding type that is selected by the plugins SCRAM-SHA-*-PLUS or GS2-KRB5-PLUS.

This solution is easy and ok for the next years, but maybe not flexible enough for future security extensions.

**Common problem**

[...]

PR #824, opened 2024-02-16
Add callback SASL_CB_SERVER_CHANNEL_BINDING

GuidoKiener:
Provide a callback function to change the channel binding type of servers (e.g. "tls-server-end-point" instead of "tls-unique") during authentication of secure mechanisms like GS2-KRB5-PLUS or SCRAM-SHA-256-PLUS.

The callback is used by the plugins SCRAM and GS2 when the desired binding type of the client does not match the binding type set with property SASL_CHANNEL_BINDING.
A server can check the requested type of channel binding and overwrite the current channel binding data with the property SASL_CHANNEL_BINDING before the authentication proceeds.

Issue #823

This patch is required to support channel binding for "tls-server-end-point" in cyrus-imapd.
See example: https://github.com/GuidoKiener/cyrus-imapd/commit/5e0af184fe8de71526bd103ecd351d70600ae863
Note that the example is not complete and requires more discussion about handling of certificates with signature algorithm SHA224, ED25519, and ED448.

### DIFF
--- a/plugins/gs2.c
+++ b/plugins/gs2.c
@@ -1025,6 +1025,25 @@ gs2_save_cbindings(context_t *text,
     return SASL_OK;
 }
 
+/*
+ * Request for (alternate) channel binding data
+ */
+static void
+gs2_server_get_channel_binding(sasl_server_params_t* sparams,
+                               const char *mech,
+                               const char *cbindingname)
+{
+    sasl_server_channel_binding_t* binding = NULL;
+    void* context = NULL;
+    /* try to get alternate channel binding data */
+    if (sparams->utils->getcallback(sparams->utils->conn,
+                                    SASL_CB_SERVER_CHANNEL_BINDING,
+                                    (sasl_callback_ft*)&binding,
+                                    &context) == SASL_OK) {
+        binding(sparams->utils->conn, context, mech, cbindingname);
+    }
+}
+
 #define CHECK_REMAIN(n)     do { if (remain < (n)) return SASL_BADPROT; } while (0)
 
 /*
@@ -1073,6 +1092,13 @@ gs2_verify_initial_message(context_t *text,
         if (ret != SASL_OK)
             return ret;
 
+        if (sparams->cbinding == NULL ||
+            strcmp(sparams->cbinding->name, text->cbindingname) != 0) {
+            /* Try to get desired channel binding data */
+            gs2_server_get_channel_binding(sparams,
+                                           text->plug.server->mech_name,
+                                           text->cbindingname);
+        }
         text->gs2_flags |= GS2_CB_FLAG_P;
         break;
     case 'n':
--- a/plugins/scram.c
+++ b/plugins/scram.c
@@ -1155,7 +1155,23 @@ cleanup:
     }
     return result;
 }
-    
+
+static void
+scram_server_get_channel_binding(sasl_server_params_t* sparams,
+				 const char *mech,
+				 const char *cbindingname)
+{
+    sasl_server_channel_binding_t* binding = NULL;
+    void* context = NULL;
+    /* try to get alternate channel binding data */
+    if (sparams->utils->getcallback(sparams->utils->conn,
+				    SASL_CB_SERVER_CHANNEL_BINDING,
+				    (sasl_callback_ft*)&binding,
+				    &context) == SASL_OK) {
+	binding(sparams->utils->conn, context, mech, cbindingname);
+    }
+}
+
 static int
 scram_server_mech_step2(server_context_t *text,
 			sasl_server_params_t *sparams,
@@ -1279,6 +1295,14 @@ scram_server_mech_step2(server_context_t *text,
                                      scram_sasl_mech);
 	    result = SASL_BADPROT;
 	    goto cleanup;
+	}
+
+	if (sparams->cbinding == NULL ||
+	    strcmp(sparams->cbinding->name, text->cbindingname) != 0) {
+	    /* Try to get desired channel binding data */
+	    scram_server_get_channel_binding(sparams,
+					     scram_sasl_mech,
+					     text->cbindingname);
 	}
 
 	if (sparams->cbinding == NULL) {
--- a/include/sasl.h
+++ b/include/sasl.h
@@ -671,6 +671,27 @@ typedef int sasl_canon_user_t(sasl_conn_t *conn,
 
 #define SASL_CB_CANON_USER (0x8007)
 
+/* Callback for servers to change channel binding type
+ * (e.g. "tls-server-end-point" instead of "tls-unique" or "tls-exporter").
+ * The callback is used by plugins like SCRAM-SHA-*-PLUS or GS2-KRB5-PLUS
+ * when the desired binding type of the client does not match the binding
+ * type set with property SASL_CHANNEL_BINDING.
+ * A server should check the requested 'cbindingname' of the 'plugin' and
+ * overwrite the channel binding data with the property SASL_CHANNEL_BINDING
+ * within the callback function.
+ * Note that plugins only store a pointer of your channel binding data and
+ * the memory of your struct sasl_channel_binding_t must be valid during
+ * the complete authentication process.
+ *
+ *  plugin        -- name of plugin
+ *  cbindingname  -- name of desired channel binding type
+ */
+typedef int sasl_server_channel_binding_t(sasl_conn_t* conn,
+					  void* context,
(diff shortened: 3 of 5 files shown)


Neustradamus:
@GuidoKiener: I only discover this PR today! :)

No three possibilities?

cc: @cyrusimap team, @aamelnikov.

GuidoKiener:
> @GuidoKiener: I only discover this PR today! :)
> 
> No three possibilities?
Which three possibilities? With the callback you have unlimited possibilities.

hyc:
The patches look OK but I think the example should be attached in a doc or README or something. The usage is not at all obvious without the example.

GuidoKiener:
> The patches look OK but I think the example should be attached in a doc or README or something. The usage is not at all obvious without the example.

I agree that an example or good README is always helpful. I also did not find a real SCRAM example in this repo with openssl and real channel binding like "tls-unique". Since I got no feedback about this patch for five months, I thought that nobody is interested in this security extension.
If you are willing to merge this branch, I can extend the simple example in https://github.com/cyrusimap/cyrus-sasl/blob/03101786b7237b5eed63ca216e2e674777f9fafa/sample/client.c#L445 and sample/server.c to show how a user can deal with two different types of channel binding. A real implementation can only be provided in cyurs-imapd (see example https://github.com/GuidoKiener/cyrus-imapd/commit/5e0af184fe8de71526bd103ecd351d70600ae863)  which is driven by future security updates and discussion.
Does it work for you when I provide this example within two weeks?

plugins/gs2.c
```diff
@@ -1073,6 +1096,13 @@ gs2_verify_initial_message(context_t *text,
         if (ret != SASL_OK)
             return ret;
 
+        if (sparams->cbinding == NULL ||
+            strcmp(sparams->cbinding->name, text->cbindingname) != 0) {
+            /* Try to get desired channel binding data */
+            gs2_server_get_channel_binding(sparams,
+                                           text->plug.server->mech_name,
+                                           text->cbindingname);
```
aamelnikov, line 1104: Overall the patch looks sensible, except for one small thing: you don't test the result of this call and you should fail authentication if it doesn't return SASL_OK.

plugins/scram.c
```diff
@@ -1281,6 +1301,14 @@ scram_server_mech_step2(server_context_t *text,
 	    goto cleanup;
 	}
 
+	if (sparams->cbinding == NULL ||
+	    strcmp(sparams->cbinding->name, text->cbindingname) != 0) {
+	    /* Try to get desired channel binding data */
+	    scram_server_get_channel_binding(sparams,
+					     scram_sasl_mech,
+					     text->cbindingname);
```
aamelnikov, line 1309: Similarly here: you should test the result of the new function call and return error if it is not SASL_OK.

plugins/gs2.c
GuidoKiener: Thanks for the review. Good point! Indeed the error handling is not smooth here and I cannot remember my initial intention.
Obviously I wanted to proceed with the default error handling of the challenge/response protocol when the client uses a wrong or unexpected channel binding type.
Currently I'm not sure if it is a good idea to abort the challenge/response protocol when the server receives an unexpected text->cbindingname. This is the current behavior of previous SASL implementations. I will think about a better solution when I add a client/server example.
What's your preference?
1. Continue with default error handling when client uses an unexpected channel binding type. If yes, I change the function declaration in static ***void*** gs2_server_get_channel_binding(...) and just ignore the callback result. You will find similar coding in lib/server.c (e.g. getopt).
2. Abort the challenge/response protocol when client uses an unexpected channel binding type or callback returns an error.

plugins/scram.c
GuidoKiener: Same here. Error handling is done in the follow-up code. Shall I change the function declaration in static ***void*** scram_server_get_channel_binding(...)

plugins/gs2.c
GuidoKiener: gs2_server_get_channel_binding now ignores the callback result. Error handling is done by protocol if callback cannot set the correct channel binding data.

plugins/scram.c
GuidoKiener: scram_server_get_channel_binding now ignores the callback result. Error handling is done by protocol if callback cannot set the correct channel binding data.

GuidoKiener:
> The patches look OK but I think the example should be attached in a doc or README or something. The usage is not at all obvious without the example.

@hyc , @quanah : A sample code is now provided. Here is an output when the client selects option -D with type 'tls-server-end-point':

```
$ ./client -D -m SCRAM-SHA-256-PLUS localhost
receiving capability list... recv: {92}
SCRAM-SHA-512-PLUS SCRAM-SHA-384-PLUS SCRAM-SHA-256-PLUS SCRAM-SHA-224-PLUS SCRAM-SHA-1-PLUS
SCRAM-SHA-512-PLUS SCRAM-SHA-384-PLUS SCRAM-SHA-256-PLUS SCRAM-SHA-224-PLUS SCRAM-SHA-1-PLUS
please enter an authentication id: guido
please enter an authorization id: guido
Password: 
send: {13}
SCRAM-SHA-256
send: {1}
Y
send: {73}
p=tls-server-end-point,a=guido,n=guido,r=FfiOkfowrbkAVZ1SpiCMVDaS+aLZDqm+
recv: {120}
r=FfiOkfowrbkAVZ1SpiCMVDaS+aLZDqm+BaiACgb6wd/Uq8A0k9P4+79mMR81ISwM,s=dcTGeYUJKRw2rzkiXZu2QM9QeDSCYDaJLb3zhHPJcR0=,i=4096
send: {184}
c=cD10bHMtc2VydmVyLWVuZC1wb2ludCxhPWd1aWRvLHVzZSBYNTA5X2RpZ2VzdCgpAA==,r=FfiOkfowrbkAVZ1SpiCMVDaS+aLZDqm+BaiACgb6wd/Uq8A0k9P4+79mMR81ISwM,p=696gLecQi1RVU+2jZOES7IA3uihIQ7wT1VEq1rc9c+Y=
recv: {46}
v=cHhVu6M81oUGC2/NnZe7AxNpUx8dssMDRPO33Xc9+3g=
send: {0}

successful authentication
closing connection
```

quanah:
Thank you very much! We will look at this as soon as we can, real life having some impact atm.

Neustradamus:
@quanah: Can you look this PR?

hyc (approved):

Neustradamus:
Thanks to @hyc for your verification!

Dear @cyrusimap team, @brong, @marclaporte, @rsto, @elliefm, @ksmurchison, @wolfsage, @dilyanpalauzov, ..

Who can confirm and merge?

@quanah has left the team, I suppose, nothing since several months.

@GuidoKiener has done a lot of work and needs to finish the support into cyrus-imapd which is linked to cyrus-sasl.
Security is very important!

Note: Debian 13 (every 2 years) will arrive soon, the new version (2.20) must be created before the freeze.
- https://github.com/cyrusimap/cyrus-sasl/milestone/5

cc: @aamelnikov.

Neustradamus:
@hyc: Thanks a lot for all your reviews, and merging! :)

@GuidoKiener: What are next steps?

GuidoKiener:
> @hyc: Thanks a lot for all your reviews, and merging! :)

@hyc: Thank you very much for merging.

> @GuidoKiener: What are next steps?

@Neustradamus:  Next steps are to revise my  [example](https://github.com/GuidoKiener/cyrus-imapd/commit/5e0af184fe8de71526bd103ecd351d70600ae863) for cyrus-imapd.
1. As noted in the first comment above the example is not complete and requires more discussion about handling of certificates with signature algorithm SHA224, ED25519, and ED448.
2. The question is also whether we should wait until the cyrus-sasl version 2.2.0 is released or if we can expect that the callback SASL_CB_SERVER_CHANNEL_BINDING is released unchanged in future versions. Then we could add support for "tls-server-end-point" before cyrus-sasl version 2.2.0 is released. However this requires to add compiler switches depending on used cyrus-sasl version.

We should continue this discussion in [cyrus-imapd #4191](https://github.com/cyrusimap/cyrus-imapd/issues/4191) and I will close this issue as soon as cyrus-sasl 2.2.0 is released.

Neustradamus:
@GuidoKiener, @hyc: I think it is needed to look asap before 2.2.0 to have a perfect version and see cyrus-imapd too.

It will be nice to have 2.2.0 before Debian 13 freeze too.

Issue #484, opened 2017-10-11
expecting time_t == unsigned long is not portable


obache:
In some source codes, it is expected that time_t == unsigned long:

https://github.com/cyrusimap/cyrus-sasl/blob/1f1c3dba7d1b727632912bb8cc9eeafca2fb5900/lib/saslutil.c#L283-L286

https://github.com/cyrusimap/cyrus-sasl/blob/1f1c3dba7d1b727632912bb8cc9eeafca2fb5900/plugins/cram.c#L138

https://github.com/cyrusimap/cyrus-sasl/blob/1f1c3dba7d1b727632912bb8cc9eeafca2fb5900/plugins/otp.c#L648-L649

https://github.com/cyrusimap/cyrus-sasl/blob/1f1c3dba7d1b727632912bb8cc9eeafca2fb5900/plugins/otp.c#L709-L710

But it is not portable, for example, 32bit arch (long=32bit) with 64bit time_t platforms,
and such mis-usage in printf/scanf cause unwanted memory access, and unwanted result.

obache:
One more usage:
https://github.com/cyrusimap/cyrus-sasl/blob/1f1c3dba7d1b727632912bb8cc9eeafca2fb5900/saslauthd/saslcache.c#L177

PR #827, opened 2024-03-20
Extend the time_t format specifiers to 64 bit

bgermann:
In some format strings, it is expected that time_t is long. long is 32 bit for 32 bit architectures, while time_t might be 64 bit. Extend the format string specifiers to long long, which is guaranteed to consist of 64 bits.

Closes: #484

### DIFF
--- a/lib/saslutil.c
+++ b/lib/saslutil.c
@@ -280,9 +280,9 @@ int sasl_mkchal(sasl_conn_t *conn,
   time(&now);
 
   if (hostflag && conn->serverFQDN)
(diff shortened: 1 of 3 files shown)


hyc:
just to note ... the right thing to do is first write an autoconf test to get the size of time_t. After that I'd define a macro for the correct printf format specifier. Also there were platforms where long = 64 and long long = 128 bits (although I think they're all extinct now, like Itanium).

bgermann:
Please see #828 for a corrected version of the patch.

PR #828, opened 2024-03-20
Extend the time_t format specifiers to long long

bgermann:
In some format strings, it is expected that time_t is the same size as long. long is 32 bit for 32 bit architectures, while time_t might be 64 bit. Extend the format string specifiers to long long, which can hold a time_t regardless of the platform and libc configuration.

Closes: #484

### DIFF
--- a/plugins/otp.c
+++ b/plugins/otp.c
@@ -639,8 +639,8 @@ static int make_secret(const sasl_utils_t *utils, const char *alg,
     bin2hex(otp, OTP_HASH_SIZE, buf);
     buf[2*OTP_HASH_SIZE] = '\0';
     
-    sprintf(data, "%s\t%04d\t%s\t%s\t%020ld",
-	    alg, seq, seed, buf, timeout);
+    sprintf(data, "%s\t%04u\t%s\t%s\t%020lld",
+	    alg, seq, seed, buf, (long long)timeout);
     
     return SASL_OK;
 }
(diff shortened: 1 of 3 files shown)


mistotebe:
@bgermann I don't think you've responded to the [review comment](https://github.com/cyrusimap/cyrus-sasl/pull/827#issuecomment-2009137030) from @hyc?

bgermann:
That is right. I have closed the previous PR because the sscanf part was wrong.

I am not going to implement the suggestion with autoconf test and macro because I do not have the time for it and I think that solution is too complex. You do not have to know the actual size of time_t to deal with it. If someone can come up with a platform example of this change being incorrect, please comment here.

This is what I have included in Debian for their time_t 64 bit transition.

plugins/otp.c
```diff
@@ -700,8 +700,10 @@ static int parse_secret(const sasl_utils_t *utils,
 	    return SASL_FAIL;
 	}
 	
-	sscanf(secret, "%s\t%04d\t%s\t%s\t%020ld",
-	       alg, seq, seed, buf, timeout);
+	long long tmptimeout;
```
hyc, line 703: Style nit - declaring variables in the middle of code is sloppy. At least, these lines should be enclosed as a block in { } brackets.

hyc (approved):

quanah:
Thank you very much!

PR #834, opened 2024-05-16
Fix some issues uncovered by a static analyzer

rcritten:
A few possible overruns and a memory leak.

### DIFF
--- a/lib/common.c
+++ b/lib/common.c
@@ -2395,18 +2395,19 @@ int _sasl_ipfromstring(const char *addr,
 
     /* Parse the address */
     for (i = 0; addr[i] != '\0' && addr[i] != ';'; i++) {
-	if (i >= NI_MAXHOST)
+	if (i >= NI_MAXHOST - 1)
 	    return SASL_BADPARAM;
 	hbuf[i] = addr[i];
     }
     hbuf[i] = '\0';
 
-    if (addr[i] == ';')
+    if (addr[i] == ';') {
 	i++;
-    /* XXX: Do we need this check? */
-    for (j = i; addr[j] != '\0'; j++)
-	if (!isdigit((int)(addr[j])))
-	    return SASL_BADPARAM;
+        /* XXX: Do we need this check? */
+        for (j = i; addr[j] != '\0'; j++)
+            if (!isdigit((int)(addr[j])))
+                return SASL_BADPARAM;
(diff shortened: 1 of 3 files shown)


lib/common.c
```diff
@@ -2395,7 +2395,7 @@ int _sasl_ipfromstring(const char *addr,
 
     /* Parse the address */
     for (i = 0; addr[i] != '\0' && addr[i] != ';'; i++) {
-	if (i >= NI_MAXHOST)
+	if (i >= NI_MAXHOST - 1)
```
mistotebe, line 2398: Doesn't this still overflow just after the loop? Sounds more like a job for strchr+memcpy...

simo5: However the following for loop also requires the overflow check.

mistotebe: > Why would it overflow? After the loop i points to the last element of the addr[] buffer

Right yeah I miscounted, but this is still pretty hard to reason about, it might be better to just pull it into the `for` test.

rcritten: It was due to the overflow check that I kept the test afterward. If the loop contains i < NI_MAXHOST then it will never detect an overflow. It seemed like a change in behavior to not return SASL_BADPARAM in this case.

hyc: That looks fine then. But while we're in this piece of code, I'd consider changing this:
```
--- a/lib/common.c
+++ b/lib/common.c
@@ -2401,12 +2401,13 @@ int _sasl_ipfromstring(const char *addr,
     }
     hbuf[i] = '\0';
 
-    if (addr[i] == ';')
+    if (addr[i] == ';') {
        i++;
-    /* XXX: Do we need this check? */
-    for (j = i; addr[j] != '\0'; j++)
-       if (!isdigit((int)(addr[j])))
-           return SASL_BADPARAM;
+       /* XXX: Do we need this check? */
+       for (j = i; addr[j] != '\0'; j++)
+           if (!isdigit((int)(addr[j])))
+               return SASL_BADPARAM;
+    }
 
     memset(&hints, 0, sizeof(hints));
     hints.ai_family = PF_UNSPEC;
````

rcritten: That makes sense. Only check that the port is numeric if one is provided.

hyc (approved):

PR #840, opened 2024-06-11
db_lmdb.c: correctly track if SASLDB_FREEPATH needs to be called

julek-wolfssl:
When the `getopt(cntxt, NULL, "sasldb_path", &p, NULL)` path is taken then SASLDB_FREEPATH is called on a value that can't be free'd and causes a segfault.


hyc:
The patch is a bit verbose, I would have just called the flag `do_freepath` but otherwise fine.

hyc (approved):

Issue #584, opened 2019-11-15
make init_mechlist thread-safe


elliefm:
Originally opened as https://github.com/cyrusimap/cyrus-imapd/issues/2813 by @vlmarek

> Hi,
> 
> We are doing some housekeeping in the changes we use in cyrus-sasl in Solaris.
> 
> https://github.com/oracle/solaris-userland/tree/master/components/cyrus-sasl/patches
> 
> And I have noticed that we have in the pipeline a fix which never got to be used. It seems simple enough, but I can't judge if this is something you would consider having. If yes, I'm happy to try to port it to current version and create a pull request. If no I will just drop it because we don't need it internally anymore.
> 
> That said I don't even see lib/client.c in your code so probably this is pointless ...
> 
> Thank you
> __
> Vlad

With the following attached patch:

https://github.com/cyrusimap/cyrus-imapd/files/3331156/116-race-condition.txt

quanah:
@vlmarek The proposed fix is in and of itself, not thread safe, as one thread could allocate and start using the mutex, and a second thread could allocate a new mutex, overwriting the pointer to the first one.  So while the problem description is correct, it needs a different solution.

simo5:
@quanah wouldn't it be on the mutexes implementation to make sure that sasl_MUTEX_ALLOC() is itself thread safe and always returns the same pointer if multiple threads are racing to call it ?

hyc:
@simo5 the API isn't designed to allow callers to pass in a unique handle, so sasl_MUTEX_ALLOC() can't know it's being called multiple times for the same purpose. It's functionally like malloc, always returning a new object on every call.

simo5:
@hyc duh right! I was looking at the tree (the specific place) and not the forest ...
I guess some trick could be done to figure out if the same handle needs to be returned based on the stack, but yeah, guess a new API would need to be provided here to do the right thing.

mistotebe:
libsasl2 isn't thread-safe until someone supplies viable callbacks via `sasl_set_mutex()`.

To get there, we might need to provide and document a single API entry point that users should call before libsasl2 is considered initialised. Probably a single one that does it for both client and server side. Getting this into 2.2 would give us a good opportunity to adjust our API and docs accordingly.

simo5:
Is there a reason why sasl2 can't use pthread_mutex by default instead of being simply thread unsafe?

mistotebe:
[...]

PR #845, opened 2024-07-22
Fix races in sasl_{client,server}_init()

mistotebe:
Should address https://github.com/cyrusimap/cyrus-sasl/issues/584#issuecomment-2019374753, @adamyi, can you confirm it works for you?

Admittedly a more complete overhaul of the `_init()`/`_close()` functions is warranted but I don't have a good approach for that yet.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -279,6 +279,7 @@ client_idle(sasl_conn_t *conn)
 int sasl_client_init(const sasl_callback_t *callbacks)
 {
   int ret;
+  void *mutex = NULL;
   const add_plugin_list_t ep_list[] = {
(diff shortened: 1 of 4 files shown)


mistotebe:
It still mutates global_callbacks, presumably in a memory-safe (even if non-deterministic) way, not sure if that's a problem. Otherwise might have to move that code around and abuse the fact that the content of global_callbacks is not touched in `common_init`.

mistotebe:
@adamyi Have you had a chance to check whether this addresses your issue or there are still scenarios where this fails for you?

Neustradamus:
@GuidoKiener: What do you think about this PR?

hyc (approved):

Neustradamus:
@hyc, @quanah: Any progress on this PR?

Issue #715, opened 2022-03-25
client_dispose is not thread safe


quanah:
Description from a Redhat patch.  Code in upstream cyrus-sasl remains unchanged.  Unclear why this was not submitted to upstream (or maybe it was and got lost?)

Description: client_dispose (lib/clinet.c) which closes a connection
of a sasl client frees mech_list if the head of the list differs
from the head of the global cmechlist->mech_list.  But there was a
possibility that the list appears in the middle of the global mech
list.  By freeing the mech, it crashed a multi-threaded sasl client.

quanah:
See also https://git.centos.org/rpms/cyrus-sasl/raw/c7/f/SOURCES/cyrus-sasl-2.1.26-make-client-thread-sage.patch

simo5:
I'll pull the new RH maintainer on this one.

hyc:
Hey thanks @simo5 but not necessary. Sorry I didn't see your comment till now.

PR #846, opened 2024-07-23
Bug 1147659 - cyrus-sasl client library (client.c) is not thread safe

rcritten:
Description: client_dispose (lib/client.c) which closes a connection of a sasl client frees mech_list if the head of the list differs from the head of the global cmechlist->mech_list.  But there was a possibility that the list appears in the middle of the global mech list.  By freeing the mech, it crashed a multi-threaded sasl client.

This patch checks each mech if it is in the global mech list or not. Only if it is not, the mech is freed.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -332,6 +332,26 @@ int sasl_client_init(const sasl_callback_t *callbacks)
   return ret;
 }
 
+/*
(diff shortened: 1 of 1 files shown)


hyc:
Thanks @rcritten but it looks like we crossed in flight. Issue #715 is already fixed.

PR #848, opened 2024-07-24
applying this old patch <https://bugs.debian.org/cgi-bin/bugreport.cg…

bAndie91:
…i?bug=689227> to make the client's remote IP address and port available in the sql plugin under the '%a' macro

### DIFF
--- a/plugins/sql.c
+++ b/plugins/sql.c
@@ -645,6 +645,7 @@ static const sql_engine_t sql_engines[] = {
 **   %p = prop
 **   %r = realm
 **   %v = value of prop
+**   %a = remote IP address;port
 **  e.g select %p from auth where user = %u and domain = %r;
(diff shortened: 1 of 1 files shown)


plugins/sql.c
```diff
@@ -668,6 +670,7 @@ static char *sql_create_statement(const char *statement, const char *prop,
     rlen = (int)strlen(realm);
     plen = (int)strlen(prop);
     vlen = (int)sql_len(value);
+    alen = (int)sql_len(remoteaddrport);
```
hyc, line 674: you also need to add at line 681 `biggest = sql_max(biggest, alen);` otherwise you can cause a buffer overflow.

hyc (approved):

hyc:
@quanah     Do a Squash and merge, thanks.

Issue #832, opened 2024-04-21
Realm appears to get added to username twice using kerberos5


iklick:
Version: 2.1.28
Mechanism: kerberos5

I start up saslauthd and then test the kerberos5 configuration with testsaslauthd without specifying the realm which runs successfully

```
testsaslauthd -u testuser -p testuserpw -s ldap
0: OK "Success."
```

When I run testsaslauthd while specifying the realm though, I get a failure

```
testsaslauthd -u testuser -p testuserpw -s ldap -r ISAACCORP
0: NO "authentication failed"
```

Looking at the saslauthd debug output, I see the values set for user has the realm tacked on now:

`saslauthd[296346] :auth failure: [user=testuser@ISAACCORP] [service=ldap] [realm=ISAACCORP] [mech=kerberos5] [reason=saslauthd internal error]`

I added my own debugging lines right before the krb5_parse_name call in saslauthd's auth_krb5.c and I see that form_principal_name has doubled up the realm.

```
    if (form_principal_name(user, service, realm, principalbuf, si
zeof (principalbuf))) {
        syslog(LOG_ERR, "auth_krb5: form_principal_name");
        return strdup("NO saslauthd principal name error");
    }

    /*
     *  ADDED BY ISAAC
     */
    printf("ISAACS DEBUG: user=%s, realm=%s\n", user, realm);
    printf("ISAACS DEBUG: principalbuf=%s\n", principalbuf);
```

The output of this custom debugging is:

```
ISAACS DEBUG: user=testuser@ISAACCORP, realm=ISAACCORP
ISAACS DEBUG: principalbuf=testuser@ISAACCORP@ISAACCORP
```

I think form_principal_name in auth_krb5.c should not add a realm if a realm has already been added to the username.

flowerysong:
This appears to be the result of passing `-r` when starting saslauthd: https://github.com/cyrusimap/cyrus-sasl/blob/ef0bafd15f5a541033c673e0e68993b07c16147a/saslauthd/saslauthd-main.c#L222-L223

This results in the realm being concatenated with the username, in addition to being passed as the realm: https://github.com/cyrusimap/cyrus-sasl/blob/ef0bafd15f5a541033c673e0e68993b07c16147a/saslauthd/saslauthd-main.c#L393-L404

[...]

PR #851, opened 2024-08-02
Fallback to realm if no domain is found in %d token expansion

andrea-reda:
If username have no domain, the %d have to use the same value of %r, so the break will be executed only if a domain is found.

Regression found in 627d84276bc16537fa70b1989047ea19aa5ab24f


quanah:
See issue #832 discussion

hyc (approved):

Neustradamus:
@hyc, @quanah: Any progress on this PR?

PR #822, opened 2024-01-02
delete build status

ryuichi1208:
It seems that Travis is no longer valid, so I will delete it.

Issue #853, opened 2024-08-03
`SASL_SEC_NONSTD_CBIND` can't be enabled?


flowerysong:
I ran across this while doing some work with the test suite, and I only have a shaky understanding of any of this code so it's possible I'm missing something.

#716 added this security flag which is supposed to allow explicitly enabling channel bindings for the GSSAPI mechanism. However, if I set this flag in the test programs it results in the GSSAPI mechanism being rejected when the library checks to see if it's suitable: https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/lib/client.c#L808-L811

When I tried adding the flag to https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/plugins/gssapi.c#L2625-L2629 and https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/plugins/gssapi.c#L1876-L1880 so that these checks could pass it appeared to revert to having channel bindings enabled by default. The only way I was able to get the tests working was to explicitly ignore this flag while doing the comparisons (https://github.com/flowerysong/cyrus-sasl/commit/1ba967884b1d8de0a3ae379f9492256513f6d59c), which feels wrong.

quanah:
hm, #862 ?

scabrero:
> hm, #862 ?

Hi @quanah,

I opened the PR and then found this issue where @flowerysong pointed out that adding the flag to `security_flags` seems to revert to having channel bindings enabled by default.

quanah:
@hyc please investigate, thanks!

GuidoKiener:
@hyc, @quanah, @simo5, @iboukris, @dilyanpalauzov 
I looked at #600, #601 and #637 and I see that current fixes do not solve all use cases for the problem to set channel binding for GSSAPI.

**Question:** This problem is now 5 years old. Does OpenLDAP still need channel binding for GSSAPI or is the problem solved in another way?

If yes, we should revert the channel binding for GSSAPI.
If no, we can do the following:

- Revert #862 
- Remove feature flags SASL_FEAT_CHANNEL_BINDING from GSSAPI and GSS-SPNEGO to avoid announcement of mechanisms like GSSAPI-PLUS or GSS-SPNEGO-PLUS.
[...]

Issue #852, opened 2024-08-03
`SASL_PATH` no longer has any effect but is still documented


flowerysong:
In 79827fbe559318553d6ccdf5ec37ae6778b4c367 (part of #745) `SASL_PLUGINDIR` replaced `SASL_PATH` as the environment variable used to configure the plugin directory at runtime. It's not clear that this was a deliberate change or what motivated it, since it's not mentioned in the commit message.

`SASL_PATH` is still referenced multiple times in the documentation, e.g. https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/docsrc/sasl/installation.rst?plain=1#L44 and https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/docsrc/sasl/quickstart.rst?plain=1#L61

It also still has a vestigial presence in `configure.ac`, though nothing references this definition: https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/configure.ac#L1391

PR #855, opened 2024-08-03
Replace Travis with GitHub Actions and fix test failures

flowerysong:
Runs the automated test suite using GHA. You can see it in action at https://github.com/flowerysong/cyrus-sasl/actions

Related: #852, #853, #854

Closes #822.
Closes #747.


hyc (changes requested):
Still waiting for the non-relevant changes to be split out of this PR.

lib/client.c
```diff
@@ -797,6 +797,9 @@ int sasl_client_start(sasl_conn_t *conn,
 
 	    myflags = conn->props.security_flags;
 
+	    /* not a real security flag */
+	    myflags &= ~SASL_SEC_NONSTD_CBIND;
```
hyc, line 801: Looks like this addresses #853 but should be in a separate PR.

lib/server.c
```diff
@@ -1315,6 +1315,9 @@ static int mech_permitted(sasl_conn_t *conn,
     /* special case plaintext */
     myflags = conn->props.security_flags;
 
+    /* not a real security flag */
+    myflags &= ~SASL_SEC_NONSTD_CBIND;
```
hyc, line 1319: As with lib/client.c, this appears to address #853

PR #866, opened 2025-01-07
update docs for newer Sphinx

rjbs:
This matches code recently merged into cyrus-imap.

1. rip out the custom theme, which was not compatible with newer Sphinx (and not worth fixing)
2. update extlinks to have exactly on %s in text part, which is required in newer Sphinx
3. update multiline uses of `c:function::`; these actually must be single-line to work properly
4. fix some other warnings from Sphinx

This does not eliminate *all* warnings.  There are still a few about duplicate definitions.  It would be nice to clean these up, but I'm not the right person at the moment.


rjbs:
I have done a test build of this branch, and the docs seem okay.

elliefm (approved):

PR #869, opened 2025-01-23
Add compatibility for gcc 15

rcritten:
Fedora 42 is going to use gcc 15 which changes some warnings into errors. Address the issues raised.

The issues addressed include:
- The RETURN macro is defined differently in two places. Rename one.
- Both atexit and the sigint and sigterm actions call server_exit(). The function arguments diff. Introduce a new generic signal handler to call server_exit() for sigint and sigterm signals.


alls0rts (approved):
Changes all look good.

hyc (approved):

PR #871, opened 2025-03-02
Provide AUTHID to GSS to select correct principal

michael-dev:
When using postfix smtp client with KRB5_CLIENT_KTNAME for relayhost authentication, the first entry in the client keytab is always used. For /etc/krb5.keytab, this might be a service principal of the host instead of the user principal, so gss_init_sec_context then fails - AS-REQ fails for service principals, only works for user principals.

Fix this by using the authid supplied by postfix to provide a desired name to gss_acquire_cred (if present).

Postfix config:
  smtp_sasl_auth_enable = yes
  smtp_sasl_mechanism_filter = gssapi
  smtp_sasl_password_maps = static:HOSTNAME$$@EXAMPLE.ORG:irrelevant
  import_environment = ... KRB5_CLIENT_KTNAME=/etc/krb5.keytab \
                           KRB5CCNAME=MEMORY:


hyc:
Why are the client and server sharing a single keytab file? That's inherently insecure, since it allows arbitrary entities to authenticate as either of them.

michael-dev:
No, client and server do not share a keytab. On the client, I have /etc/krb5.keytab which contains the machine account. So I want to use the machine account of the client to authenticate to the server (for system mails, e.g.from CRON).
Similiarly, I want to be able to use a gMSA account to authenticate to the mail gateway (for application e-Mail). The current key of the gMSA account would automatically be provided using a keytab file.

Neustradamus:
@hyc, @quanah: Have you progressed on this PR?

PR #873, opened 2025-04-09
Fix build with gcc-15 error in cyrus-sasl-2.1

markyang92:
Fix build with gcc-15 error in cyrus-sasl-2.1


hyc (commented):
This patch to saslauthd/md5.c is ok for 2.1 but the code no longer exists in master/2.2.

PR #876, opened 2025-07-20
Update GSSAPI max SSF in docs

Freddo3000:
This PR updates the documentation to reflect https://github.com/cyrusimap/cyrus-sasl/pull/429 which increases max SSF from 56 to 256.

On another note, there are no 56-bit supported encryption methods following kerberos 1.18. `K5_MIN_SSF` might need to be updated to reflect that.


Neustradamus:
@hyc, @quanah: Have you seen this PR?

PR #877, opened 2025-08-07
Use HMAC-SHA256 for cache passwords over MD5

Firstyear:
Currently the password cache uses MD5 for password caching in the mmapped cache file. This is not cryptographically secure.

This changes the algorithm to HMAC-SHA256. While other password verification algorithms are considered cryptographically superior (such as argon2id) in this case we want to maintain the cache's high performance, while improving the security of cached passwords.

To achieve this, the HMAC key itself is ephemeral and only ever stored in memory. This means that per invocation of saslauthd the HMAC key will be randomised. This results in the cache mmap file being effectively useless to an attacker who manages to steal the cache as they lack the HMAC key to attempt to validate or bruteforce any cached password.

This also means that if saslauthd crashes, the key material to access the cache is lost, effectively invalidating all records in that cache.

If an attacker were able to access the HMAC key in memory, then we can pretty safely assume that the attacker can also access plaintext password material making the need to access the key irrelevant.

As a result, this change improves security of cached passwords without a significant loss of performance in the cache's operation.

### DIFF
--- a/saslauthd/cache.c
+++ b/saslauthd/cache.c
@@ -55,6 +55,8 @@
 #include "utils.h"
 #include "globals.h"
 #include <openssl/evp.h>
+#include <openssl/rand.h>
+#include <openssl/hmac.h>
 
 /****************************************
  * module globals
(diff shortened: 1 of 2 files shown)


Neustradamus:
@hyc, @quanah: Have you seen this PR?

PR #886, opened 2026-04-22
Fix: Include <time.h> unconditionally in saslutil.c for GCC 14+ compatibility

### LINKED ISSUE #859: 2.1.28: Issues with including time.h
Hi! :wave: 

Upon rebuilding the libsasl component for Arch Linux I noticed that we have issues with the (missing) inclusion of `time.h` now.

I believe the following should check for `HAVE_SYS_TIME_H` instead:
https://github.com/cyrusimap/cyrus-sasl/blob/537af16986caff73ef45ab54a5a020bfb636ddaf/lib/saslutil.c#L62-L64

This issue is also present in the (now deleted) plugins/cram.c.

I'll have to backport the fix for the deleted file now, but I can open a PR for the fixing of saslutil.c.
That said: A new release would be very good to have.

jamesbirrellgray:
### Description
This PR adopts a robust pattern for including time headers in `lib/saslutil.c`. It ensures `<time.h>` is included unconditionally while maintaining the legacy conditional inclusion for `<sys/time.h>`.

### Motivation
On modern distributions using **GCC 14** (such as **Ubuntu 24.04**), the build fails with a fatal error:
`error: implicit declaration of function ‘clock’ [-Wimplicit-function-declaration]`

The existing logic wraps `<time.h>` in a `HAVE_TIME_H` check, which can fail to be defined correctly during the configuration phase in certain environments. Moving the include outside the conditional block ensures that the `clock()` function is always declared.

### Impact
This bug is a significant blocker for the **Homebrew on Linux** ecosystem. Its failure prevents the installation of **R (r-base)** and critical R packages (e.g., `curl`, `httr`) that rely on `cyrus-sasl` for authentication foundations.

### Related Issues
Fixes #859: https://github.com


thebaptiste:
I think `#include <time.h>` is also missing in plugin/cram.c

Neustradamus:
@thebaptiste: You have specified cram.c, I think there is an error, it is scram.c?

@jamesbirrellgray: Good catch, thanks!

thebaptiste:
> @thebaptiste: You have specified cram.c, I think there is an error, it is scram.c?
> 
No, it was cram.c but I was building the last release 2.1.28. cram.c seems to have been removed since.
A new release (building ok on gcc 14) would be great (2.1.28 is 4 years old)

